PR #5, opened 2019-04-30
Narration sentence highlighting (BL-7024)

sujeffreyl:
Supports sentence-by-sentence highlighting of an audio file corresponding to a whole text box
(Confer https://github.com/BloomBooks/BloomDesktop/pull/3158)

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/5)

### DIFF
--- a/src/bloom-player.less
+++ b/src/bloom-player.less
@@ -22,7 +22,8 @@ above the page so that page numbers will increment.*/
     background-color: darkgray;
 }
 
-.ui-audioCurrent {
+span.ui-audioCurrent:not(.disableHighlight),
+div.ui-audioCurrent:not(.disableHighlight) p {
     background-color: yellow;
 }
 
--- a/src/narration.ts
+++ b/src/narration.ts
@@ -26,6 +26,7 @@ export default class Narration {
     private elementsToPlayConsecutivelyStack: HTMLElement[];
     private nextElementIdToPlay: string;
     private timingsStack: number[];
+    private elementsToHighlightStack: HTMLElement[];
 
     public PageNarrationComplete: LiteEvent<HTMLElement>;
     public PageDurationAvailable: LiteEvent<HTMLElement>;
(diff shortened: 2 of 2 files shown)


sujeffreyl:
A good test URL: url=https://s3.amazonaws.com/BloomLibraryBooks-Sandbox/jeffrey_su@sil.org/91183f55-56fb-463b-bbb3-e619260027dd/AutoSplit+Timings+4.6

sujeffreyl:
___
*[src/narration.ts, line 78 at r3](https://reviewable.io/reviews/bloombooks/bloom-player/5#-LeJ3naG1s5o2jp0IQVV:-LeY51uQCN96McL422ks:bym2e4l) ([raw file](https://github.com/bloombooks/bloom-player/blob/b3ef7931e058ca34eead5c7c80685c17ecbf7699/src/narration.ts#L78)):*

We have a convention...which sadly I violated with ui-audio and friends...that classes which mean something to the bloom code should start with "bloom-", like bloom-translationGroup. So probably this should be bloom-sentence. I also wonder whether bloom-segment would be better...it's not certain that we will continue forever to break paragraphs up strictly into sentences. bloom-audioHighlightSegment (or bloom-audioHighlightChunk?) would be clearer still, though I'm not sure it's worth the extra length. @hatton, what do you think?
</blockquote></details>

maybe .bloom-highlightSegment?

sujeffreyl:
___
*[src/narration.ts, line 103 at r3](https://reviewable.io/reviews/bloombooks/bloom-player/5#-LeJALX7-b3F4v-Ny061:-Lew3MyE3Nt1NjzsT1RU:b-f02mst) ([raw file](https://github.com/bloombooks/bloom-player/blob/b3ef7931e058ca34eead5c7c80685c17ecbf7699/src/narration.ts#L103)):*

I checked documentation and experimentally confirmed as well. Promise is fulfilled when play is STARTED successfully.
Source: https://developers.google.com/web/updates/2016/03/play-returns-promise
</blockquote></details>

Updating the comment...

PR #18, opened 2019-05-31
BL-7211 Add app bar

JohnThomson:
- with back button and play/pause
- also: got rid of black bar at bottom of page
- center player when too narrow to fill window
- add material design and theme

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/18)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "bloom-player",
-    "version": "0.9.0",
+    "version": "0.9.1",
     "description": "A tool for displaying Bloom books in iframes or WebViews",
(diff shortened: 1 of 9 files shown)


JohnThomson:
@hatton, this change makes bloomPlayer.js more than ten times larger. Should we investigate that before merging?

hatton:
@JohnThomson 1) increase % doesn't matter, rather, absolute size increase, right? 2) was that a dev build or a production build (with tree shaking)?

JohnThomson:
I _thought_ the production build was using tree-shaking, but apparently not. Working on changing that...

PR #40, opened 2019-07-26
BL-7443 move more analytics handling into player

JohnThomson:
...and various other changes connected with using in BloomReader, such as an alternative way to communicate with the outer tool.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/40)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "bloom-player",
-    "version": "0.9.23",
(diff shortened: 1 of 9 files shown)


JohnThomson:
This has had an extensive interactive review. Doesn't need more, but don't merge yet, as a few more changes are coming.

PR #42, opened 2019-07-31
BL-7423 Load images progressively

hatton:
Video and audio are loaded as needed. Images are loaded in the background, with no limits at this point (e.g. not just one page ahead).

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/42)


andrew-polk:
Oops. Merged this just before realizing it needed to bump the version number. Doing so now.

PR #70, opened 2019-10-02
BL-7610 Add language chooser

gmartin7:
* language data provided by callback from
   bloom-player-core to bloom-player-controls
* LanguageMenu sends new selected language
   up to BloomPlayerControls where the language
   data is stored and change is sent down to
   BloomPlayerCore via the activeLanguageCode prop
* for now we are getting language iso codes and
   display names from 'language-display-names'
   field of meta.json
* fix bug in fixRelativeUrls()
* this is a second try at a PR; this one rebuilds the Swiper
   when the language changes

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/70)

### DIFF
--- a/src/bloom-player-controls.tsx
+++ b/src/bloom-player-controls.tsx
@@ -15,6 +15,7 @@ import { ControlBar } from "./controlBar";
 import { ThemeProvider } from "@material-ui/styles";
 import theme from "./bloomPlayerTheme";
 import React, { useState, useEffect } from "react";
+import LangData from "./langData";
 
 // This component is designed to wrap a BloomPlayer with some controls
 // for things like pausing audio and motion, hiding and showing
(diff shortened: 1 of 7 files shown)


andrew-polk:
___
*[src/bloom-player-core.tsx, line 540 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/70#-LqNtOm_AXWX35GTL3xc:-LqbNPbQ2jwruTLPEGKA:b-he3fps) ([raw file](https://github.com/bloombooks/bloom-player/blob/360aad7295dfce5f110d5f3dc074f624f08d985f/src/bloom-player-core.tsx#L540)):*
<details><summary><i>Previously, gmartin7 (Gordon Martin) wrote…</i></summary><blockquote>

I occurs to me that if we switch to only skipping xmatter divs, books that are designed to show 2 or 3 languages won't anymore in `BP`.
</blockquote></details>

I'm pretty sure your current code only skips divs in xmatter since it only skips divs with the data-book attribute.

PR #84, opened 2019-10-31
Fix BR display/scrolling problems with old books (BL-7773)

StephenMcConnel:
Although not the initial problem reported, this problem was discovered
when testing this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/84)


StephenMcConnel:
We've decided not to pursue this fix.  It's too risky for the possible benefit.

PR #119, opened 2020-05-05
Another try at safely shrinking margins on A5 pages

JohnThomson:
This version only does it when the marginBox really does have the expected size.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/119)


andrew-polk:
On hold indefinitely.

PR #146, opened 2020-09-01
Send analytics directly from bloom-player (BL-8874)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/146)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,22 @@ Example: `hideFullScreenButton=true`
 
 Default: `false`
 
+#### independent
+
+If true, bloom-player reports its own analytics directly to segment.  If false, bloom-player sends messages to its host, and the host is responsible for reporting analytics.
+
+Example: `independent=false`
+
+Default: `true`
+
+#### host
+
+If set, provides a host value for analytics.  If not set, bloom-player attempts to derive a value from the available information if independent is `true` (information can be limited in an iframe) and does nothing otherwise.
+
+Examples: `host=bloomlibrary` or `host=bloomreader` or `host=bloompubviewer` or `host=readingapp` or `host=embed@xyzzy.org`
+
+Default: nothing/undefined ??
+
 # Development
 
 Run `yarn` to get the dependencies.
--- a/index-for-developing.html
+++ b/index-for-developing.html
@@ -3,16 +3,15 @@
 <html>
     <body>
         <iframe
-        disabled_local_src="bloomplayer-for-developing.htm?url=http://localhost:8080/test_books/MoonAndCap/MoonAndCap.htm&hideFullScreenButton=false"
+        local_src="bloomplayer-for-developing.htm?url=http://localhost:8080/test_books/MoonAndCap/MoonAndCap.htm&hideFullScreenButton=false&independent=true&host=testing"
         src="bloomplayer-for-developing.htm?url=https://s3.amazonaws.com/bloomharvest/educationforlife%40sil.org%2f433f2ae9-3cba-48bf-82a7-faa3932908ae/bloomdigital%2findex.htm&hideFullScreenButton=false"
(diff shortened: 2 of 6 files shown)


StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/146)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloomPlayerAnalytics.tsx, line 8 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/146#-MGE4iBY6nqp_UrugSsD:-MGJfR3UCSL_zIAnqHm7:b-471q76) ([raw file](https://github.com/bloombooks/bloom-player/blob/e95720b47ba4e4ce81b57732a3cd944cc39dd348/src/bloomPlayerAnalytics.tsx#L8)):*

Technically (should be changed on BL too) also changed to load a modified analytics.min.js that uses our proxy instead of sending direct to segment.io.
</blockquote></details>

Done.  Added extra item to comment.
___
*[src/bloomPlayerAnalytics.tsx, line 133 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/146#-MGEACTQ2kYsLlKIdQmp:-MGJh9ft8ppaAvyH7Jl7:b46076d) ([raw file](https://github.com/bloombooks/bloom-player/blob/e95720b47ba4e4ce81b57732a3cd944cc39dd348/src/bloomPlayerAnalytics.tsx#L133)):*

I think it would be helpful to explain, here or in finalAnalytics, that it's not certain...maybe even not likely?...that finalAnalytics actually manages to send the report, and that's why we want to report here.
</blockquote></details>

Done.  I added some commentary in both places.
___
*[src/bloomPlayerAnalytics.tsx, line 180 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/146#-MGE5Ei8DvGqPQ32nhMv:-MGJhIHcCIiMKur3bgBu:b-71w3qp) ([raw file](https://github.com/bloombooks/bloom-player/blob/e95720b47ba4e4ce81b57732a3cd944cc39dd348/src/bloomPlayerAnalytics.tsx#L180)):*

I haven't reviewed this file in detail as I assumed it's copied from BL. You might want to draw my attention to anything you had to change that I should look at.
</blockquote></details>

Maybe the biggest change was to delete the useTrack method.  I couldn't see any need for it in this context since we aren't showing the analytics to the user anywhere.
___
*[src/bookInfo.ts, line 107 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/146#-MGEBzOW-XpB8Onaashw:-MGJho6U1hwumwkK3NdT:b70g75n) ([raw file](https://github.com/bloombooks/bloom-player/blob/e95720b47ba4e4ce81b57732a3cd944cc39dd348/src/bookInfo.ts#L107)):*

I guess none of our other ideas for figuring out the host worked out? e.g., we talked about referrer, I think
</blockquote></details>

There may be some work remaining for dealing with embedded use that might use the referrer window.parent.location value.  My mind is still very hazy on how exactly embedding is supposed to work.  But for those hosts that are fixed in nature, and which we can change, setting the host parameter is far and away the easiest and most reliable mechanism.  (It may be that way too for embedding, except I'm not sure how much control we have over the url parameters in that case.)

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/146#-:-MGhwZP94WpIrc_pqwmV:bnfp4nl)* status: 0 of 6 files reviewed, all discussions resolved

JohnThomson (approved):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/146)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #147, opened 2020-09-01
Fix bug where pause/play causes highlighting to stop working (BL-8906)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/147)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -1368,9 +1368,10 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
 
     // called by narration.ts
     public static storeAudioAnalytics(duration: number): void {
-        if (duration < 0.001) {
+        if (duration < 0.001 || Number.isNaN(duration)) {
             return;
         }
+
         const player = BloomPlayerCore.currentPagePlayer;
         player.bookInteraction.totalAudioDuration += duration;
 
--- a/src/narration.ts
+++ b/src/narration.ts
@@ -40,9 +40,9 @@ export default class Narration {
     public PageDurationAvailable: LiteEvent<HTMLElement>;
     public PageDuration: number;
 
-    // This represents the original time when the audio started. Even if the user presses pause/play, it doesn't change.
+    // A Session Number that keeps track of each time playAllSentences started.
     // This is used to determine whether the page has been changed or not.
-    private audioPlayOriginalStartTime: number | null = null; // milliseconds (since 1970/01/01, from new Date().getTime())
+    private currentAudioSessionNum: number = 0;
 
     // This represents the start time of the current playing of the audio. If the user presses pause/play, it will be reset.
     // This is used for analytics reporting purposes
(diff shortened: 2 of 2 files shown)


sujeffreyl:
___
*[src/narration.ts, line 629 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/147#-MGDzzVUEuwBwk29doIM:-MGE3kI04Zygj068ZwbO:bdcohvr) ([raw file](https://github.com/bloombooks/bloom-player/blob/051d160d288512aeb12306cd171e1958e8f202cc/src/narration.ts#L629)):*

Doesn't seem to be a new problem, but let's say the sound is 5s long and they already played 2s. Did we already report that 2s when they paused? If not, and if audioPlayCurrentStartTime is used to figure out the duration to report, we'll only report 3s of duration when the rest of the sound finishes.
</blockquote></details>

Yes, existing behavior is that we report 2s when they hit pause, and 3 s when it finishes. That's what I read from the code and also confirmed experimentally.

When I made this change, I just assumed the existing behavior was the desired behavior, just wrote it so as to preserve the existing behavior.

andrew-polk (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/147)* status: all files reviewed, 5 unresolved discussions (waiting on @JohnThomson and @sujeffreyl)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/147)* status: all files reviewed, 4 unresolved discussions (waiting on @sujeffreyl)

___
*[src/narration.ts, line 45 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/147#-MGE0xXS86cBoHPz7FQM:-MGG29KwB7pGI7LiOb8E:b-du2e6y) ([raw file](https://github.com/bloombooks/bloom-player/blob/051d160d288512aeb12306cd171e1958e8f202cc/src/narration.ts#L45)):*

Would this allow for the best of both? 
Call it something like currentlyPlayingAudioSession{Id} and make it a guid?
That's less efficient, though...
Otherwise, maybe most of the problem can be alleviated with a new name, perhaps something like currentlyPlayingAudioSession?
</blockquote></details>

I don't object to using the time as the value, just to a name that suggests that what time it is matters. I take Jeffrey's point that 'id' is getting overloaded...in fact it has a similar chance of confusion to time because it suggests that it is the element ID of the thing being played. Sorry, my bad. currentAudioSessionMarker? currentPlaybackMarker? playbackSessionIndicator? {current}audioPlaySequence?

andrew-polk (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/147)* status: all files reviewed, 4 unresolved discussions (waiting on @sujeffreyl)

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/147)* status: all files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[src/narration.ts, line 45 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/147#-MGE0xXS86cBoHPz7FQM:-MGJk5pPEs_OLWQXaCPC:b5a911e) ([raw file](https://github.com/bloombooks/bloom-player/blob/051d160d288512aeb12306cd171e1958e8f202cc/src/narration.ts#L45)):*

I don't object to using the time as the value, just to a name that suggests that what time it is matters. I take Jeffrey's point that 'id' is getting overloaded...in fact it has a similar chance of confusion to time because it suggests that it is the element ID of the thing being played. Sorry, my bad. currentAudioSessionMarker? currentPlaybackMarker? playbackSessionIndicator? {current}audioPlaySequence?
</blockquote></details>

how about currentAudioSessionNum?  I can get behind the idea of a session number.

PR #156, opened 2020-09-30
BloomPlayer now expects URLs to be properly encoded (BL-9041)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/156)

### DIFF
--- a/src/bloom-player-controls.tsx
+++ b/src/bloom-player-controls.tsx
@@ -28,7 +28,11 @@ import {
 // for testing the BloomPlayer narration functions.
 
 interface IProps {
-    unencodedUrl: string; // url of the bloom book (folder)
+    // Url of the bloom book (folder). Should be a valid, well-formed URL
+    // e.g. any special chars in the path or book title should be appropriately encoded using URL (percent) encoding
+    // e.g. if title is C#, url should be http://localhost:8089/bloom/C:/PathToTemp/PlaceForStagingBook/C%23
+    // (Not "[...]/C#" nor "[...]%2FC%2523", and most certainly not with any XML (entity) encoding)
+    url: string;
     initiallyShowAppBar: boolean;
     allowToggleAppBar: boolean;
     showBackButton: boolean;
@@ -200,11 +204,8 @@ export const BloomPlayerControls: React.FunctionComponent<IProps &
         let localMaxPageDimension = maxPageDimension;
         let localAspectRatio = pageAspectRatio;
         const pageClass = BloomPlayerCore.getPageSizeClass(page);
-        if (
-            props.unencodedUrl !== previousUrl ||
-            pageClass !== previousPageClass
-        ) {
-            setPreviousUrl(props.unencodedUrl);
+        if (props.url !== previousUrl || pageClass !== previousPageClass) {
+            setPreviousUrl(props.url);
             setPreviousPageClass(pageClass);
             // Some other one-time stuff:
             // Arrange for this to keep being called when the window size changes.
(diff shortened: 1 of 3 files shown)


andrew-polk (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/156)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[src/bloom-player-core.tsx, line 1468 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/156#-MIV6Kx-9bC1N3ARt1EX:-MIV6Kx-9bC1N3ARt1EY:b-ivnnri) ([raw file](https://github.com/bloombooks/bloom-player/blob/eba916bb01f558c76faa3874cc21e03cd989afe5/src/bloom-player-core.tsx#L1468)):*
I'm finding this in the original issue which added this:
> I made a change to bloompub-reader to go looking for htms as a last resort, like BR does. But that leaves a problem that maybe we aren't being disciplined about how we encode/decode the "url" parameter in the querystring given to bloom-player. I think the name is decieving because this "url" is actually not the url, it's a paramter in the querystring of a larger url. So if I'm correct, it should be encoded/decoded using endodeuricomponent & decodeuricomponent.
>
>LATER: I did code that change up, but it turned out not to be the problem, so I erred on the side of caution and did not submit that change.
>
>Instead, I just fixed the problem if "#" in the file name or path name, because that, alone, was being sent through un-encoded by axios.get.

So I'm thinking we might need a change in BloomPubReader before we can do this?

andrew-polk:
___
*[src/bloom-player-core.tsx, line 1468 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/156#-MIV6Kx-9bC1N3ARt1EX:-MIV6YxUEFEBu4WomRAi:b-5r3mey) ([raw file](https://github.com/bloombooks/bloom-player/blob/eba916bb01f558c76faa3874cc21e03cd989afe5/src/bloom-player-core.tsx#L1468)):*

I'm finding this in the original issue which added this:
> I made a change to bloompub-reader to go looking for htms as a last resort, like BR does. But that leaves a problem that maybe we aren't being disciplined about how we encode/decode the "url" parameter in the querystring given to bloom-player. I think the name is decieving because this "url" is actually not the url, it's a paramter in the querystring of a larger url. So if I'm correct, it should be encoded/decoded using endodeuricomponent & decodeuricomponent.
>
>LATER: I did code that change up, but it turned out not to be the problem, so I erred on the side of caution and did not submit that change.
>
>Instead, I just fixed the problem if "#" in the file name or path name, because that, alone, was being sent through un-encoded by axios.get.

So I'm thinking we might need a change in BloomPubReader before we can do this?
</blockquote></details>

https://issues.bloomlibrary.org/youtrack/issue/BL-8652

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/156)* status: 1 of 3 files reviewed, all discussions resolved

___
*[src/bloom-player-core.tsx, line 1468 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/156#-MIV6Kx-9bC1N3ARt1EX:-MIWV0AR2yDTsncHDhTU:b-sm2pco) ([raw file](https://github.com/bloombooks/bloom-player/blob/eba916bb01f558c76faa3874cc21e03cd989afe5/src/bloom-player-core.tsx#L1468)):*

https://issues.bloomlibrary.org/youtrack/issue/BL-8652
</blockquote></details>

Jeffrey dealt with this.

andrew-polk (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/156)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #169, opened 2020-12-02
Add OverlayScrollbars to handle overflowing text boxes instead (BL-9249)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/169)

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,8 @@
         "testEnvironment": "jsdom"
     },
     "dependencies": {
-        "dynamic-import-polyfill": "^0.1.1"
+        "@types/overlayscrollbars": "1.12.0",
+        "dynamic-import-polyfill": "^0.1.1",
+        "overlayscrollbars": "1.13.0"
     }
 }
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -40,6 +40,8 @@ import { LegacyQuestionHandler } from "./activities/legacyQuizHandling/LegacyQui
 import { CircularProgress } from "@material-ui/core";
 import { BookInfo } from "./bookInfo";
 import { BookInteraction } from "./bookInteraction";
+import OverlayScrollbars from "overlayscrollbars";
+import "./overlayScrollbars-bloom.css"; // The CSS for the OverlayScrollbars plugin. This is a modified version for bloom-player that fixes CSS conflicts due to specificity.
 
 // BloomPlayer takes a URL param that directs it to Bloom book.
 // (See comment on sourceUrl for exactly how.)
@@ -1147,6 +1149,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                 this.swiperInstance = s;
             },
             simulateTouch: true, //Swiper will accept mouse events like touch events (click and drag to change slides)
+            touchStartPreventDefault: false,    // If true, would prevent the default, which would cause overlayScrollbars not to receive mousedown events.
 
             on: {
                 slideChange: () => {
@@ -1455,6 +1458,19 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                 this.swiperInstance.keyboard.disable();
             } else {
                 this.swiperInstance.keyboard.enable();
+            }
+
+            const options = OverlayScrollbars.defaultOptions();
+
+            // Ignore the cover page, the title one doesn't work well..
+            // ENHANCE: Figure out why it sets the title height to 0 when you attach OverlayScrolbars to it.
+            if (!bloomPage.classList.contains("cover")) {
+                // Ignore editables that are under a bloom-textOverPicture... This isn't working too well for them right now.
+                // The ignored ones are comic bubbles.  Comic bubbles usually shouldn't need a scrollbar because the size auto-grows in Bloom,
+                // so the only way you're supposed to be able to get an overflow is if the there is if the bubble can't grow anymore
+                // (because it's already reached the bottom of the image container).
+                const editables = bloomPage.querySelectorAll(":not(.bloom-textOverPicture) > .bloom-translationGroup .bloom-editable.bloom-visibility-code-on");
+                OverlayScrollbars(editables, options);
             }
         }, 0); // do this on the next cycle, so we don't block scrolling and display of the next page
     }
--- a/src/bloom-player.less
+++ b/src/bloom-player.less
@@ -212,6 +212,18 @@ We just make them follow the main content normally. */
         }
     }
 
+    // Make it so the text isn't selectable
+    // (This was added because when we set swiper's touchStartPreventDefault to false,
+    // the clicks were allowed to go through again and a double-click on text was capable of selecting it)
+    .bloom-editable {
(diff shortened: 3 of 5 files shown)


JohnThomson (changes requested):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/169)* status: all files reviewed, 2 unresolved discussions (waiting on @sujeffreyl)

___
*[src/bloom-player.less, line 226 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/169#-MNdRmkF4AxrOQqnYBJR:-MNdRmkF4AxrOQqnYBJS:bmw7qkl) ([raw file](https://github.com/bloombooks/bloom-player/blob/f979b691ff4c6d451c32903d245f72eb039a37cf/src/bloom-player.less#L226)):*
> ```Less
>         user-select:none;
>     }
> ```

If the problem is  only double-clicks, does that really matter? Doesn't seem very likely someone will do a double-click by accident. I can imagine it being quite annoying to be reading a book and unable to select and copy something I want to quote. Probably not an issue for people in the early stages of literacy, but might be for a more sophisticated user (e.g., proof-reading).
Probably a question for @hatton if you didn't already run this by him.
___
*[src/bloom-player-core.tsx, line 1472 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/169#-MNdQUgqEg4be2gUO-as:-MNdQUgr5TurDXBtOoGK:b-4843ty) ([raw file](https://github.com/bloombooks/bloom-player/blob/f979b691ff4c6d451c32903d245f72eb039a37cf/src/bloom-player-core.tsx#L1472)):*
> ```XML
>                 // so the only way you're supposed to be able to get an overflow is if the there is if the bubble can't grow anymore
>                 // (because it's already reached the bottom of the image container).
>                 const editables = bloomPage.querySelectorAll(":not(.bloom-textOverPicture) > .bloom-translationGroup .bloom-editable.bloom-visibility-code-on");
> ```

Feels like this could be very inefficient. Unless the browser is very smart, it will start by finding almost everything (except .bloom-textOverPicture) and then search all the subtrees for bloom-translationGroups, and then search all those subtrees for visible boom editables. Is it possible to test how long this takes on a good-sized book?
Feel free to push back if this is premature optimization...I don't know for a fact that it will be a problem...but seems it would be more efficient to do one search for visible bloom-editables, and then filter out any that aren't in translation groups (probably none?) or that are in bloom-textOverPictures.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/169)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @sujeffreyl)

___
*[src/bloom-player.less, line 226 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/169#-MNdRmkF4AxrOQqnYBJR:-MNdYeeP6P2LSk16y6V8:boy0wi1) ([raw file](https://github.com/bloombooks/bloom-player/blob/f979b691ff4c6d451c32903d245f72eb039a37cf/src/bloom-player.less#L226)):*

If the problem is  only double-clicks, does that really matter? Doesn't seem very likely someone will do a double-click by accident. I can imagine it being quite annoying to be reading a book and unable to select and copy something I want to quote. Probably not an issue for people in the early stages of literacy, but might be for a more sophisticated user (e.g., proof-reading).
Probably a question for @hatton if you didn't already run this by him.
</blockquote></details>

Yeah, these accidental selections do happen everywhere we don't explicitly defend against them, and they are annoying.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/169)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson and @sujeffreyl)

___
*[src/bloom-player.less, line 226 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/169#-MNdRmkF4AxrOQqnYBJR:-MNdYU_aCL4wxmxYtc-j:b-i05d6d) ([raw file](https://github.com/bloombooks/bloom-player/blob/f979b691ff4c6d451c32903d245f72eb039a37cf/src/bloom-player.less#L226)):*

Yeah, these accidental selections do happen everywhere we don't explicitly defend against them, and they are annoying.
</blockquote></details>

I tried it out again when allow double-clicking. It's not very discoverable, and even if you do, it's an awful experience. I tried to select just "hi" from "This page ...". I couldn't manage it. Also, it only works if you double-click, then drag down. (If you double-click and drag sideways, swiper activates first). So yeah, IMO it'd be better to just disable it. If we need it, we'll need to implement it for real.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/169)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 1472 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/169#-MNdQUgqEg4be2gUO-as:-MNe0_GfBQxJRI8kMnR7:bcuy1nz) ([raw file](https://github.com/bloombooks/bloom-player/blob/f979b691ff4c6d451c32903d245f72eb039a37cf/src/bloom-player-core.tsx#L1472)):*

Feels like this could be very inefficient. Unless the browser is very smart, it will start by finding almost everything (except .bloom-textOverPicture) and then search all the subtrees for bloom-translationGroups, and then search all those subtrees for visible boom editables. Is it possible to test how long this takes on a good-sized book?
Feel free to push back if this is premature optimization...I don't know for a fact that it will be a problem...but seems it would be more efficient to do one search for visible bloom-editables, and then filter out any that aren't in translation groups (probably none?) or that are in bloom-textOverPictures.
</blockquote></details>

Summary: Did some research. Conclusion - Won't Fix.

Long answer...
1) There is merit to your intuition. Originally I had just the ".bloom-editable.bloom-visibility-code-on".  When it's written like that (eliminating 2 levels), it is noticeably faster. When eliminating only the :not(...) part, it is perhaps marginally faster... the one with 2 levels instead of 3 was faster 58% of the time. (Not too convincing)

2) Test on how long it takes on a good-sized book: It's only scanning from the page, so I'd say a more relevant test is to test on Normal/commonplace pages than on big books.

3) Premature optimization? Yes: It was normally taking < 0.04 ms. (That is, 4% of a millisecond.) 

4) The cure of then filtering out the bad ones seemed worse than the disease. So, then I added a forEach loop, a call to .closest inside each loop, and also have array manipulation going on. I tested this against Iteration 1 in 4 books (the original Chetana book where this repro'd, my test book, Moon & Cap on dev, and one of the IMS books on dev). Iteration 1 was faster 18-0, 7-0, 13-1, and 29-2. Generally roughly 2x faster than the filtering approach.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/169)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #188, opened 2021-01-26
The fix seems to be needed also on Safari (BL-9304)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -658,8 +658,16 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                 replayButton
             );
         }
-        if (navigator.userAgent.includes("Chrome")) {
-            // due to bug in Chrome, we can't be sure button will show up if we place it over the video
+        // from https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent
+        const UA = navigator.userAgent;
+        const isWebkit =
+            /\b(iPad|iPhone|iPod)\b/.test(UA) &&
+            /WebKit/.test(UA) &&
+            !/Edge/.test(UA) &&
+            !window.MSStream;
+        if (isWebkit || /chrome)/.test(UA.toLowerCase())) {
+            // Due to a bug in Chrome and Safari (and probably other Webkit-based browsers, hopefully
+            // this catches them all), we can't be sure the button will show up if we place it over the video
             // So, instead, we hide the video to make room for it. That can seem a very abrupt change,
             // so we fade the video out before replacing it with the button.
             parent.insertBefore(replayButton, pageVideoData.video); // there but still display:none


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzhOlf9HuYs0LwiPIz:-MRzhOlf9HuYs0LwiPJ-:b-b5l755) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*
> ```XML
>             );
>         }
>         if (/(safari|chrome)/.test(navigator.userAgent.toLowerCase())) {
> ```

So I'm not clear about the selection here. Chrome is no longer "webkit-based" unless it's on IOS. On desktop, it is "blink" / chromium (which historically derived from webkit). Edge is also chromium-based.  So.... I'm not clear from this is what you want?

sujeffreyl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzi18kEhjZBZc_Uvec:-MRzi18kEhjZBZc_Uved:b-bwkqvr) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*
> ```
> (safari
> ```
Alternatively, for testing for Webkit browsers, I created a function isMacOrIOS() in osUtils.ts.  (Apparently Chrome on iOS need to use webkit basis too). I investigated using User Agent string, but with user agent strings saying all kinds of things, there were some Android phones that were false positives.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzhOlf9HuYs0LwiPIz:-MS-9ACm8CBCKssGTz0o:bmpbeog) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*

So I'm not clear about the selection here. Chrome is no longer "webkit-based" unless it's on IOS. On desktop, it is "blink" / chromium (which historically derived from webkit). Edge is also chromium-based.  So.... I'm not clear from this is what you want?
</blockquote></details>

It seems to be clearly the same bug in Safari, whatever the history of the browser engines: the button disappears except when it's in the upper left quadrant. Colin says Edge is already working right, the remaining problems (at least of things he tried) are Safari and Chrome/IOS. The easiest way to test seems to be to push it out. It's not very harmful if we get it wrong. If we do it for a browser where we don't need the fix, it just means the video is hidden when the button shows though it didn't need to be. If we don't do it for some obscure browser where it's needed until we find out about it, well, the new feature will sometimes be missing on that browser.
This bug doesn't seem to have been reported so there's no way to be sure which browsers will have it. And from what I've read, the whole user agent thing is very flaky, with different versions of different browsers pretending to be something else to try to work around pages which misuse it to try to guess features. I don't know how to get this perfectly right. Maybe we should just stop trying and hide the video when the button is showing on every platform?

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzhOlf9HuYs0LwiPIz:-MS-K9ct5m5bAr4C5KhA:b-n2zmgl) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*

It seems to be clearly the same bug in Safari, whatever the history of the browser engines: the button disappears except when it's in the upper left quadrant. Colin says Edge is already working right, the remaining problems (at least of things he tried) are Safari and Chrome/IOS. The easiest way to test seems to be to push it out. It's not very harmful if we get it wrong. If we do it for a browser where we don't need the fix, it just means the video is hidden when the button shows though it didn't need to be. If we don't do it for some obscure browser where it's needed until we find out about it, well, the new feature will sometimes be missing on that browser.
This bug doesn't seem to have been reported so there's no way to be sure which browsers will have it. And from what I've read, the whole user agent thing is very flaky, with different versions of different browsers pretending to be something else to try to work around pages which misuse it to try to guess features. I don't know how to get this perfectly right. Maybe we should just stop trying and hide the video when the button is showing on every platform?
</blockquote></details>

You seem confirm what I'm saying, that this a webkit problem (safari & Chrome/ios, presumably firefox ios as well), *not* a chrome problem. Yet the code in question is asking if it is Chrome.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzhOlf9HuYs0LwiPIz:-MS-V5puB3325WjDaGeB:bvfkvsx) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*

You seem confirm what I'm saying, that this a webkit problem (safari & Chrome/ios, presumably firefox ios as well), *not* a chrome problem. Yet the code in question is asking if it is Chrome.
</blockquote></details>

Well, you say Chrome is NOT webkit-based any more, which agrees with my own research, and Chrome definitely DOES need the fix. So maybe we're looking at Chrome OR webkit? I'll see if I can figure out an implementation of that.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzhOlf9HuYs0LwiPIz:-MS0T1WQD-W0YxYE4_yy:bkau3oh) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*

Well, you say Chrome is NOT webkit-based any more, which agrees with my own research, and Chrome definitely DOES need the fix. So maybe we're looking at Chrome OR webkit? I'll see if I can figure out an implementation of that.
</blockquote></details>

Sorry, I got stuck in some mental rut. So we want ios|safari|chrome, and this is the closest this userAgent call can come to that?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @hatton and @JohnThomson)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzhOlf9HuYs0LwiPIz:-MS3QK6T-rRTi-k1q8SJ:b97m5z8) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*

Sorry, I got stuck in some mental rut. So we want ios|safari|chrome, and this is the closest this userAgent call can come to that?
</blockquote></details>

I had another go. Hopefully this is more likely to catch all the ones that need it and not too many that don't. If testing still discovers problems, I think we should just go to replacing the video with the button everywhere.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @hatton and @sujeffreyl)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzi18kEhjZBZc_Uvec:-MSU0nJI3t-PWinwE2k2:b-88w7iw) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*

> ```
> (safari
> ```
Alternatively, for testing for Webkit browsers, I created a function isMacOrIOS() in osUtils.ts.  (Apparently Chrome on iOS need to use webkit basis too). I investigated using User Agent string, but with user agent strings saying all kinds of things, there were some Android phones that were false positives.
</blockquote></details>

The problem is definitely more than just MacOrIOS; it happens in Chrome on Windows, at least. My latest patch is already following guidelines for catching webkit-based browsers plus Chrome (which apparently was originally based on Webkit but forked).
As I said before, if this version still misses any browsers which have the problem, I think we should give up and do the fix everywhere. If there are a few that get the fix and don't need it, I'm inclined to live with that.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/188)* status: all files reviewed, 2 unresolved discussions (waiting on @hatton and @sujeffreyl)

___
*[src/bloom-player-core.tsx, line 661 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/188#-MRzi18kEhjZBZc_Uvec:-MSUKPQkAP1wx9NqwC7u:b-vs6jde) ([raw file](https://github.com/bloombooks/bloom-player/blob/37ec85f6792dddd200a622ed9e17bf59a2ad7127/src/bloom-player-core.tsx#L661)):*

The problem is definitely more than just MacOrIOS; it happens in Chrome on Windows, at least. My latest patch is already following guidelines for catching webkit-based browsers plus Chrome (which apparently was originally based on Webkit but forked).
As I said before, if this version still misses any browsers which have the problem, I think we should give up and do the fix everywhere. If there are a few that get the fix and don't need it, I'm inclined to live with that.
</blockquote></details>

I think the suggestion from Jeffrey would be to do 
if (isMacOrIOS() || /chrome/)

If that doesn't seem better, please go ahead and merge.

PR #190, opened 2021-02-02
Fix display bug with activity pages in publish preview (BL-8900)

StephenMcConnel:
I tried this in storybook in both firefox 60 and chromium 88.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/190)

### DIFF
--- a/src/bloom-player.less
+++ b/src/bloom-player.less
@@ -585,3 +585,12 @@ body,
 *[tabindex]:focus {
     outline: none;
 }
+
+// FF60 has a behavior that we don't see in new FF or other browsers (BL-8900), where a following activity page
+// seems to grow and push into the current page. We don't fully understand what causes this, but the following
+// collapses the width of the following page, so that this does not happen.
+// See https://developer.mozilla.org/en-US/docs/Web/CSS/visibility for an explanation of what this setting does.
+// Note that swiper uses flex boxes to achieve its behavior.
+.swiper-slide-next {
+    visibility: collapse;
+}


hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/190)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[src/bloom-player.less, line 589 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/190#-MSc_YHU1t_egrY7dPds:-MSc_YHU1t_egrY7dPdt:bgykuq) ([raw file](https://github.com/bloombooks/bloom-player/blob/368f36c3e7165cb370ec2deb9f750fe59c062a0c/src/bloom-player.less#L589)):*
> ```Less
> }
> 
> // BL-8900: This is needed for Firefox60/Geckofx60 to get around an wierd bug: a following activity page intrudes itself
> ```

After talking, here's a comment we could have here

FF60 has a behavior that we don't see in new FF or other browsers (BL-8900), where an activity seems to grow and push into the current page. We don't fully understand what causes this, but the following collapses the width of the following page, so that this does not happen.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/190)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[src/bloom-player.less, line 589 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/190#-MSc_YHU1t_egrY7dPds:-MScfrcE8M_cUagPKQ6h:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/368f36c3e7165cb370ec2deb9f750fe59c062a0c/src/bloom-player.less#L589)):*

After talking, here's a comment we could have here

FF60 has a behavior that we don't see in new FF or other browsers (BL-8900), where an activity seems to grow and push into the current page. We don't fully understand what causes this, but the following collapses the width of the following page, so that this does not happen.
</blockquote></details>

Done.

andrew-polk (approved):

PR #194, opened 2021-02-22
Ensure state.styleRules does not get modified after initial pass (BL-9504)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/194)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -122,7 +122,12 @@ interface IProps {
 }
 interface IState {
     pages: string[]; // of the book. First and last are empty in context mode.
-    styleRules: string; // concatenated stylesheets the book references or embeds.
+
+    // concatenated stylesheets the book references or embeds.
+    // Make sure these are only set once per book or else
+    // it wreaks havoc on scoped styles. See BL-9504.
+    styleRules: string;
+
     importedBodyClasses: string;
     // indicates current page, though typically not corresponding to the page
     // numbers actually on the page. This is an index into pages, and in context
@@ -151,8 +156,6 @@ interface IState {
     // If the user interacts, which we detect as anything that makes us change page,
     // we're no longer in a FORCED pause, though we may still be paused.
     inPauseForced: boolean;
-
-    usingDefaultLang: boolean;
 }
 
 export class BloomPlayerCore extends React.Component<IProps, IState> {
@@ -188,8 +191,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
         loadErrorHtml: "",
         ignorePhonyClick: false,
         isFinishUpForNewBookComplete: false,
-        inPauseForced: false,
-        usingDefaultLang: true
+        inPauseForced: false
     };
 
     // The book url we were passed as a URL param.
@@ -489,6 +491,12 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
             this.bookInteraction.clearPagesShown();
             this.music.processAllMusicForBook(pages);
         }
+
+        const preferredLanguages = this.bookInfo.getPreferredTranslationLanguages();
+        const usingDefaultLang =
+            preferredLanguages[0] === this.props.activeLanguageCode ||
+            !this.props.activeLanguageCode;
+
         for (let i = 0; i < pages.length; i++) {
             const page = pages[i] as HTMLElement;
             const landscape = this.setPageSizeClass(page);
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
Couple of minor suggestions.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/194#-:-MUFMliJEFWFPgylZ794:b-bxheso)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/bloom-player-core.tsx, line 559 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/194#-MUFJHGa89SUnbPaW9or:-MUFJHGa89SUnbPaW9os:b-bpi6st) ([raw file](https://github.com/bloombooks/bloom-player/blob/275b45625f5e8abc198c7295c66ba9759152532e/src/bloom-player-core.tsx#L559)):*
> ```XML
>         }
> 
>         // Make sure you don't modify state.styleRules after the initial pass.
> ```

I don't think 'initial pass' is quite clear enough. Because of the await, the state.styleRules IS being modified after the first call to finishUp in the first pass through componentDidUpdate. And it could get modified again if we somehow got a new book, though I'm not sure any of our clients does that.
Is the essence of it that styleRules must not be changed once isLoading is set false so the book actually renders?
I don't suppose there's any way to encapsulate styleRules so this mistake CAN'T be made? (I can't think of a good one, but maybe you can?)
Possibly also worth commenting where styleRules is declared.
___
*[src/bloom-player-core.tsx, line 633 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/194#-MUFIoQw4ZuRj4qc51Mx:-MUFIoQw4ZuRj4qc51My:bd9zpvg) ([raw file](https://github.com/bloombooks/bloom-player/blob/275b45625f5e8abc198c7295c66ba9759152532e/src/bloom-player-core.tsx#L633)):*
> ```XML
>     // We decided (BL-9256) that Title-On-Cover should not display in the national language
>     // unless we are displaying the book's default language or unless the national language IS
>     // the main one we're showing (that is, it has bloom-content1 as well as bloom-contentNational1).
> ```

So, the essence of the change is that this code, which previously modified the style definition, now modifies a class on the element, right?
In view of all your work, might be worth a comment along the lines "don't be tempted to achieve this by modifying the style rules..."

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/194)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 559 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/194#-MUFJHGa89SUnbPaW9or:-MUFtUcNCoyvGbxC9ZJ3:b-5sser0) ([raw file](https://github.com/bloombooks/bloom-player/blob/275b45625f5e8abc198c7295c66ba9759152532e/src/bloom-player-core.tsx#L559)):*

I don't think 'initial pass' is quite clear enough. Because of the await, the state.styleRules IS being modified after the first call to finishUp in the first pass through componentDidUpdate. And it could get modified again if we somehow got a new book, though I'm not sure any of our clients does that.
Is the essence of it that styleRules must not be changed once isLoading is set false so the book actually renders?
I don't suppose there's any way to encapsulate styleRules so this mistake CAN'T be made? (I can't think of a good one, but maybe you can?)
Possibly also worth commenting where styleRules is declared.
</blockquote></details>

I've attempted to clarify, but I'm not sure I've really addressed your question.
The essence of it is if we modify the actual styles once they have gotten set (if we call setState with a different string), the polyfill gets confused.
___
*[src/bloom-player-core.tsx, line 633 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/194#-MUFIoQw4ZuRj4qc51Mx:-MUFtloPCsaOrFGmFM6O:b24ue9z) ([raw file](https://github.com/bloombooks/bloom-player/blob/275b45625f5e8abc198c7295c66ba9759152532e/src/bloom-player-core.tsx#L633)):*
> So, the essence of the change is that this code, which previously modified the style definition, now modifies a class on the element, right?

Right.

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/194)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

PR #199, opened 2021-03-24
Hack to repair FF60 scroll bug (BL8900)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199)


andrew-polk (changes requested):
> Reviewed 92 of 92 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199)* status: all files reviewed, 6 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player.less, line 334 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaPGTV5abiJip-kQ6a:-MWaPGTV5abiJip-kQ6b:b-pgzi3v) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player.less#L334)):*
> ```Less
> // .moving .swiper-slide-next .bloomPlayer-page {
> //     display: block;
> // }
> ```

Any reason to keep these?
___
*[src/bloom-player-core.tsx, line 251 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaVXYT8uFkAVLqF0gv:-MWaVXYT8uFkAVLqF0gw:bkg5pvj) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L251)):*
> ```XML
>     }
> 
>     private enableFF60Repair = 0;
> ```

I'm having a little bit of a hard time with the naming/typing of this variable.
One solution would be to have two variables, one with the time and another with a boolean. Then you would need a function to toggle enabled/disabled.
But maybe that's overkill....

Maybe sufficient would be to move this variable declaration after the hack comment below, and then expand it to say explicitly how this variable is used (positive means enabled, vice versa; and we progressively reduce the number while checking for the bug).
___
*[src/bloom-player-core.tsx, line 259 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaToK-0gkA9w42cSPs:-MWaToK-0gkA9w42cSPt:b-mci6kq) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L259)):*
> ```
>     // Whatever goes wrong seems to happen within a couple of seconds of the
>     // page before the problem one stabilizing (probably, when the widget gets loaded for real as the
>     // 'next page').
> ```
I had a hard time parsing this sentence at first.
Would this be better?

The errant shift seems to happen within a couple of seconds of the page-before-the-widget-page becoming the current page (probably, when the widget gets lazy-loaded as the 'next page').

But my main hang-up was the phrase, "page before the problem one stabilizing," so maybe that's all that needs to change.
___
*[src/bloom-player-core.tsx, line 286 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaWrwa3sLYLIzFk7uT:-MWaWrwa3sLYLIzFk7uU:bwy1xhz) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L286)):*
> ```XML
>         this.enableFF60Repair -= 100;
>         setTimeout(() => this.repairFF60Offset(), 100);
>     }
> ```

Would it be worth wrapping the whole contents of this function in a try/catch?
___
*[src/bloom-player-core.tsx, line 1532 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaR_UE-XtXZeTE30yX:-MWaR_UE-XtXZeTE30yY:bm21rds) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L1532)):*
> ```
> 2s of checking seems to be enough to fix problems
> ```
I'm dubious. If this is, as you speculate, related to the amount of time it takes to load the "next page," that could easily be more than 2 seconds when downloading an activity over the internet (assuming that's when the download happens).
___
*[src/stories/index.tsx, line 168 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaXkPHCw-dJPb8RRB0:-MWaXkPI1FjiXItUrLV0:bck13e1) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/stories/index.tsx#L168)):*
> ```XML
>     "activity scroll issue",
>     //"https://s3.amazonaws.com/bloomharvest-sandbox/stephen_mcconnel%40sil.org/eda23196-72e8-4701-8651-58f14ce2bcfd/bloomdigital/index.htm"
>     "src/Testing Away Again/index.htm"
> ```

Can we move this out to a directory in the root (sibling of src) called "Test Books" or similar?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199)* status: 0 of 92 files reviewed, 6 unresolved discussions (waiting on @andrew-polk)

___
*[src/bloom-player.less, line 334 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaPGTV5abiJip-kQ6a:-MWe9rBO-IFw24DiP8wt:b-i01wcs) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player.less#L334)):*

Any reason to keep these?
</blockquote></details>

No. Done.
___
*[src/bloom-player-core.tsx, line 251 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaVXYT8uFkAVLqF0gv:-MWe9uRsCenxw2AICpWU:b-7j6lp0) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L251)):*

I'm having a little bit of a hard time with the naming/typing of this variable.
One solution would be to have two variables, one with the time and another with a boolean. Then you would need a function to toggle enabled/disabled.
But maybe that's overkill....

Maybe sufficient would be to move this variable declaration after the hack comment below, and then expand it to say explicitly how this variable is used (positive means enabled, vice versa; and we progressively reduce the number while checking for the bug).
</blockquote></details>

It used to be boolean and should have been renamed. Sorry.
___
*[src/bloom-player-core.tsx, line 259 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaToK-0gkA9w42cSPs:-MWeFMjb1TP3JF5YDnVq:bxk3q0x) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L259)):*

> ```
>     // Whatever goes wrong seems to happen within a couple of seconds of the
>     // page before the problem one stabilizing (probably, when the widget gets loaded for real as the
>     // 'next page').
> ```
I had a hard time parsing this sentence at first.
Would this be better?

The errant shift seems to happen within a couple of seconds of the page-before-the-widget-page becoming the current page (probably, when the widget gets lazy-loaded as the 'next page').

But my main hang-up was the phrase, "page before the problem one stabilizing," so maybe that's all that needs to change.
</blockquote></details>

Rewrote.
___
*[src/bloom-player-core.tsx, line 286 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaWrwa3sLYLIzFk7uT:-MWeFmZO1rq7SAqhXPhX:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L286)):*

Would it be worth wrapping the whole contents of this function in a try/catch?
</blockquote></details>

Done.
___
*[src/bloom-player-core.tsx, line 1532 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaR_UE-XtXZeTE30yX:-MWeFoyVC4TX7HJzeZYU:b-cec1q0) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L1532)):*

> ```
> 2s of checking seems to be enough to fix problems
> ```
I'm dubious. If this is, as you speculate, related to the amount of time it takes to load the "next page," that could easily be more than 2 seconds when downloading an activity over the internet (assuming that's when the download happens).
</blockquote></details>

I would hope BR activities would be designed to work offline. But I agree, it should be increased just in case. Done. (As a result, I found I needed another turn-it-off event handler to make sure it doesn't interfere with dragging.)
___
*[src/stories/index.tsx, line 168 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaXkPHCw-dJPb8RRB0:-MWeKvPy-DqSDP52pbJF:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/stories/index.tsx#L168)):*

Can we move this out to a directory in the root (sibling of src) called "Test Books" or similar?
</blockquote></details>

Done.

andrew-polk (changes requested):
> Reviewed 2 of 93 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199)* status: 2 of 92 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[src/bloom-player-core.tsx, line 1532 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaR_UE-XtXZeTE30yX:-MWeWWl-0SWm90xOwkRm:bo66wdn) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L1532)):*

I would hope BR activities would be designed to work offline. But I agree, it should be increased just in case. Done. (As a result, I found I needed another turn-it-off event handler to make sure it doesn't interfere with dragging.)
</blockquote></details>

I'm not sure I understand your comments about activities not using the internet.
The use case I'm thinking of is reading a book on bloomlibrary.org where we expect the browser to ask for the activity's files when the activity is the next page.

Speaking of "next page," I guess you tried swiping from the end (after hitting the end key) such that it loads when it is the previous page?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199)* status: 2 of 92 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/bloom-player-core.tsx, line 1532 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaR_UE-XtXZeTE30yX:-MWesw126JxDFNvJ63Ul:b-xsae61) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L1532)):*

I'm not sure I understand your comments about activities not using the internet.
The use case I'm thinking of is reading a book on bloomlibrary.org where we expect the browser to ask for the activity's files when the activity is the next page.

Speaking of "next page," I guess you tried swiping from the end (after hitting the end key) such that it loads when it is the previous page?
</blockquote></details>

I hadn't tried the end key. Don't think I even knew we handled it! But it seems fine.
You're right that an activity (or any page) will typically use the internet while the book is read online. So I've adjusted the comment.
My thought was that a book shouldn't use the internet in such a way that it needs things that are not PART of the book...so, in BloomReader, it shouldn't need the internet, and likewise, when previewing in Bloom Editor it shouldn't need it.
Apart from someone who has somehow managed to hang on to a copy of FireFox 60 despite all Mozilla's efforts in the past couple of years, or has a browser with the same bug, it's only someone doing a preview in Bloom Editor that needs the activity to load within the time limit. And THAT is what I was thinking shouldn't be affected by a slow internet.

andrew-polk (approved):
> Reviewed 91 of 93 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 1532 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/199#-MWaR_UE-XtXZeTE30yX:-MWf7D2-2sU6srq7CZX_:b-jty38d) ([raw file](https://github.com/bloombooks/bloom-player/blob/05e89e6b59fd6bff5add61c1e815a7ad0e7d2142/src/bloom-player-core.tsx#L1532)):*

I hadn't tried the end key. Don't think I even knew we handled it! But it seems fine.
You're right that an activity (or any page) will typically use the internet while the book is read online. So I've adjusted the comment.
My thought was that a book shouldn't use the internet in such a way that it needs things that are not PART of the book...so, in BloomReader, it shouldn't need the internet, and likewise, when previewing in Bloom Editor it shouldn't need it.
Apart from someone who has somehow managed to hang on to a copy of FireFox 60 despite all Mozilla's efforts in the past couple of years, or has a browser with the same bug, it's only someone doing a preview in Bloom Editor that needs the activity to load within the time limit. And THAT is what I was thinking shouldn't be affected by a slow internet.
</blockquote></details>

Yes, my fault for not remembering this is specific to FF60. Thanks.

StephenMcConnel (approved):
As "horrible hacks" go, this does appear to work and to be suitably constrained.  I suppose you could check navigator.userAgent to check for FF60 specifically, but that would probably be overkill.  :lgtm:

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199#-:-MWkcWi8884KpIxXo01P:b-nephyu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

andrew-polk (approved):
If we could reliably detect FF60, I actually think it would be worth it. Because then we don't have the polling happening.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199#-:-MWkoFGN5MK2-SAvjz0k:buwfqi5)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

andrew-polk (approved):
Technically, I think we actually pass in the host, so we could even base it on being inside BloomDesktop. We probably don't know how recently the bug was fixed in FF...

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199#-:-MWko_fx4b43c5kQYddh:b-hdqs3h)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

JohnThomson (commented):
I don't feel very guilty if things misbehave because someone, presumably by their own choice, is using an old version of a browser. It actually feels pretty safe to say we only do this when running from Bloom desktop. So that's well worth considering.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199#-:-MWxs6wM0TSbKG-JIZ7v:b-m1cmvp)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

JohnThomson (commented):
OK, now the hack only happens when host=bloomdesktop

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199#-:-MWxzBDcATXmW6Tv4OPE:baq394q)* status: 91 of 92 files reviewed, all discussions resolved (waiting on @andrew-polk)

andrew-polk (approved):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/199)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #201, opened 2021-04-01
Integrate bloom-player localization with Crowdin (BL-9134)

StephenMcConnel:
This doesn't try to upload the English messages.json to Crowdin, but
everything else is automated to work at build time if an API token is
available via the process environment.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/201)


andrew-polk (changes requested):
> Reviewed 6 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/201)* status: 6 of 18 files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[webpack.common.js, line 84 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYI0ao0VhOcil8Xnwg:-MXYI0apEFSx46TIqKOq:bvffgj4) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L84)):*
> ```JavaScript
>                     .finally(() => {
>                         CombineTranslatedMessagesFiles();
>                         callback(); // don't call it if you do want to stop compilation
> ```

I'm pretty dubious about swallowing errors (here and below).
It seems like we have two scenarios in which this code will run:
1. We have the env var set. Then I think we want to be alerted about any failures. I.e. the build should fail.
2. We do not have the env var set. Then we skip all this code (as you've already written it).

So then you can get rid of the finally and else and just call the Combine... and callback unconditionally (and let the throw be uncaught or whatever is the right thing to fail the build).

I'm also not convinced there is any benefit to having localized strings in scenario 2. Then we wouldn't need the non-English files checked in at all, right?

Open to discussion or 3rd opinion.
___
*[webpack.common.js, line 234 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYFg1A70T_NvIUMdJZ:-MXYFg1A70T_NvIUMdJ_:b-sbjbue) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L234)):*
> ```JavaScript
>             targetLanguageId: lang,
>             fileIds: [kCrowdinBloomPlayerFileId],
>             // REVIEW: should we set exportApprovedOnly: true ?
> ```

Yes, I think so. That's what BloomDesktop does, right?
(with the caveat that the user can turn on unapproved)
___
*[webpack.common.js, line 235 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYJfr_EVUs3MONoZEh:-MXYJfr_EVUs3MONoZEi:brezmm1) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L235)):*
> ```JavaScript
>             fileIds: [kCrowdinBloomPlayerFileId],
>             // REVIEW: should we set exportApprovedOnly: true ?
>             skipUntranslatedStrings: true
> ```

Is there a way to make it not download the descriptions?
___
*[webpack.common.js, line 309 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYHZ0C3F2DsF9c56ye:-MXYHZ0C3F2DsF9c56yf:bxy1hxp) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L309)):*
> ```JavaScript
>             return fetchCrowdinData(translationsApi, lang);
>         });
>         return Promise.all(downloads);
> ```

Maybe I'm just reading the code wrong, but it seems like you handle all the fetch/saves before the final ones in this method and then handle the final saves outside this function. Wouldn't it simplify things to handle all fetch/saves here?

All this got me wondering if we can just download the files directly rather than creating them from the data.
I see that in BloomLibrary2, crowdin-sync.ts, that is what we do. But we have to do a "build" first. Did you consider that approach? I'm not sure what the tradeoffs would be, but one advantage is that we have code already there doing it that way.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/201)* status: 6 of 18 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[webpack.common.js, line 84 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYI0ao0VhOcil8Xnwg:-MXYTFZQ69Trja5HH6M-:b1o55zk) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L84)):*

I'm pretty dubious about swallowing errors (here and below).
It seems like we have two scenarios in which this code will run:
1. We have the env var set. Then I think we want to be alerted about any failures. I.e. the build should fail.
2. We do not have the env var set. Then we skip all this code (as you've already written it).

So then you can get rid of the finally and else and just call the Combine... and callback unconditionally (and let the throw be uncaught or whatever is the right thing to fail the build).

I'm also not convinced there is any benefit to having localized strings in scenario 2. Then we wouldn't need the non-English files checked in at all, right?

Open to discussion or 3rd opinion.
</blockquote></details>

I think the Combine has to be called in either a then or finally clause so that the build effectively waits for the message file to be constructed.
Allowing the throw to be uncaught may be good enough to stop the build in its tracks.
___
*[webpack.common.js, line 234 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYFg1A70T_NvIUMdJZ:-MXYZv128CfWgDKCqAAN:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L234)):*

Yes, I think so. That's what BloomDesktop does, right?
(with the caveat that the user can turn on unapproved)
</blockquote></details>

Done.
___
*[webpack.common.js, line 235 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYJfr_EVUs3MONoZEh:-MXY_q3t1wP2ZzNb02Ow:b-yovsi9) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L235)):*

Is there a way to make it not download the descriptions?
</blockquote></details>

I don't see anything in the API documentation to do this.
___
*[webpack.common.js, line 309 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYHZ0C3F2DsF9c56ye:-MXYjWN-2hwIsIcu4jhm:b-fogcny) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L309)):*

Maybe I'm just reading the code wrong, but it seems like you handle all the fetch/saves before the final ones in this method and then handle the final saves outside this function. Wouldn't it simplify things to handle all fetch/saves here?

All this got me wondering if we can just download the files directly rather than creating them from the data.
I see that in BloomLibrary2, crowdin-sync.ts, that is what we do. But we have to do a "build" first. Did you consider that approach? I'm not sure what the tradeoffs would be, but one advantage is that we have code already there doing it that way.
</blockquote></details>

Since bloom-player has only a single file to download (and is very unlikely to ever get any more l10n message files), I thought using the exportProjectTranslation API for that single file would be simpler that using a approach like BloomLibrary2 (which I did look at).
Crowdin imposes a maximum of 20 concurrent accesses to the API for any one account.  We aren't pushing that at the moment for bloom-player, but there are over 30 different languages at least started in the SIL-Bloom project, so we do have to account for this limit.  I've simplified the code to fetch all the information in the one loop in this method, and to save all the files in the calling method in a single forEach statement.

andrew-polk (approved):
> Reviewed 1 of 18 files at r1, 1 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/201)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[webpack.common.js, line 84 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/201#-MXYI0ao0VhOcil8Xnwg:-MXZ8SQt31gWNSztMtSY:btkvgez) ([raw file](https://github.com/bloombooks/bloom-player/blob/14aabbee8bc76ed52079c5af069eb8ce736d29c3/webpack.common.js#L84)):*

I think the Combine has to be called in either a then or finally clause so that the build effectively waits for the message file to be constructed.
Allowing the throw to be uncaught may be good enough to stop the build in its tracks.
</blockquote></details>

`await` is another option.
I'll let you decide if that is worth what I think would be a simplification while I get the env var set up on the build agent.

PR #223, opened 2021-07-15
BL-9871 Allow widgets to "own" nav fullscreen

gmartin7:
* add Active Presenter widget to storybook
* use 'data-player-options' on bloom-page to pass options
   to BP

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/223)


andrew-polk (changes requested):
> Reviewed 28 of 28 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/223)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[src/activities/iframeActivity.ts, line 28 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/223#-Mel5t5V2mZIRzrzGJuw:-Mel5t5V2mZIRzrzGJux:b-80101i) ([raw file](https://github.com/bloombooks/bloom-player/blob/f2b0bc8aebfa7070bbd1eda27fbceaec868a9463/src/activities/iframeActivity.ts#L28)):*
> ```XML
>         if (context.pageOptions && context.pageOptions.hideNavigation) {
>             context.hidePageNavigationButtons();
>         }
> ```

Hm. I wasn't expecting this implementation. I was expecting the logic to be separate from activities.

Pros of letting core find and process the attribute: 
- Though I can't think of a use case, there is a theoretical case that non-activity pages may want to hide navigation controls.
- Future options could easily not be related to activities.
- By waiting until the activity `start` gets called and then sending a message, I think we are still subject to flashing. My guess is that letting the core find the attribute and process it would eliminate the flash.

Thoughts?

gmartin7 (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/223)* status: 23 of 28 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/activities/iframeActivity.ts, line 28 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/223#-Mel5t5V2mZIRzrzGJuw:-MezwdUV2ZgoKTmt0sgo:b-x0lsie) ([raw file](https://github.com/bloombooks/bloom-player/blob/f2b0bc8aebfa7070bbd1eda27fbceaec868a9463/src/activities/iframeActivity.ts#L28)):*

Hm. I wasn't expecting this implementation. I was expecting the logic to be separate from activities.

Pros of letting core find and process the attribute: 
- Though I can't think of a use case, there is a theoretical case that non-activity pages may want to hide navigation controls.
- Future options could easily not be related to activities.
- By waiting until the activity `start` gets called and then sending a message, I think we are still subject to flashing. My guess is that letting the core find the attribute and process it would eliminate the flash.

Thoughts?
</blockquote></details>

* YAGNI
* True. This new version lets us get any page options as soon as the page is showing (since the page loads before it's visible, onload is too soon).
* So far so good vis a vis flashing...
* Had to switch to single quotes that we replace with double quotes before parsing, since Bloom escapes double quotes when saving.

andrew-polk (approved):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/223)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gmartin7)

PR #225, opened 2021-07-26
Fix crash in Chrome on iOS (BL-10216)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/225)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -1543,7 +1543,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
             console.log(
                 "getPlayerOptionsForPage failed to parse json: " +
                     optionJson +
-                    " with errror " +
+                    " with error " +
                     e.message
             );
             return;
--- a/src/externalContext.ts
+++ b/src/externalContext.ts
@@ -27,8 +27,11 @@ export function sendMessageToHost(messageObj: WithMessageType) {
     }
     // If we're in a native iOS WKWebView, we could not get it to receive postMessage
     // messages, so instead it has to inject an object called webkit.messageHandlers to receive the message.
-    if ((window as any).webkit) {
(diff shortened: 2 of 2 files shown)


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/225)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[src/externalContext.ts, line 30 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/225#-MfYY0pa1MTeafEBi2bo:-MfYY0pa1MTeafEBi2bp:b3rixl8) ([raw file](https://github.com/bloombooks/bloom-player/blob/7a03c9358231746171196e85fc9175943ebee8ca/src/externalContext.ts#L30)):*
> ```XML
>     // If we're in a native iOS WKWebView, we could not get it to receive postMessage
>     // messages, so instead it has to inject an object called webkit.messageHandlers to receive the message.
>     // Note: receiveMessage is custom defined. It may or may not exist, even if messageHandlers is added. See BL-10216
> ```

I made it such that if receiveMessage is not defined, it continues executing the rest of this function, rather than an immediate return. My rationale is that before Chris added the webkit stuff, it would've tried executing the part of the function below this, so what I implemented is better align with the legacy path.

Let me know if you think the other way is better.

JohnThomson (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/225)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/225)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

PR #230, opened 2021-08-09
Allow blorg to report readDuration (BL-10230)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/230)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,9 +88,11 @@ Default: `true`
 
 If set, provides a host value for analytics.  If not set, bloom-player attempts to derive a value from the available information if independent is `true` (information can be limited in an iframe) and does nothing otherwise.
 
-Examples: `host=bloomlibrary` or `host=bloomreader` or `host=bloompubviewer` or `host=readingapp` or `host=embed@xyzzy.org`
+Examples: `host=bloomlibrary` or `host=bloomreader` or `host=bloompubviewer` or `host=readerapp`
 
-Default: nothing/undefined ??
+...or, theoretically, `host=embed@xyzzy.org` (but we haven't used that yet)
+
+Default: `nothing/undefined`
 
 # Development
 
--- a/src/externalContext.ts
+++ b/src/externalContext.ts
@@ -1,5 +1,8 @@
 import { TransientPageDataSingleton } from "./transientPageData";
-import { getBooleanUrlParam } from "./utilities/urlUtils";
+import {
+    getBooleanUrlParam,
+    getQueryStringParamAndUnencode
+} from "./utilities/urlUtils";
 
 import {
     track,
@@ -81,11 +84,17 @@ export function updateBookProgressReport(event: string, properties: any) {
     if (getBooleanUrlParam("independent", true)) {
         updateBookProgress(event, { ...ambientAnalyticsProps, ...properties });
     } else {
-        // The host must handle this property itself because there is no accurate
+        // Non-web hosts must handle this property themselves because there is no accurate
         // way for bloom-player to accumulate time without knowing when the host
         // is in the foreground.
         // i.e. window.blur and window.focus do not get called in Android Webview.
-        delete properties.readDuration;
+        //
+        // We decided to whitelist the hosts which receive this property because
+        // we want to be safe. We would rather report no data than incorrect/invalid data.
+        // We could try to do something fancy like looking for parent.window to distinguish
+        // between running on the web or not, but that is not guaranteed to be future-proof, either.
+        if (getQueryStringParamAndUnencode("host", "") !== "bloomlibrary")
+            delete properties.readDuration;
 
         sendMessageToHost({
             messageType: "updateBookProgressReport",


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/230)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/externalContext.ts, line 92 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/230#-MglJB_0E9OsqsnNi-pL:-MglJB_0E9OsqsnNi-pM:bnn9f4e) ([raw file](https://github.com/bloombooks/bloom-player/blob/ea6da1779114cfcb70898a6a91bf9f74fdd44422/src/externalContext.ts#L92)):*
> ```XML
>         // i.e. window.blur and window.focus do not get called in Android Webview.
>         if (getQueryStringParamAndUnencode("host", "") !== "bloomlibrary")
>             delete properties.readDuration;
> ```

I don't see why we ever delete it. If the host is a web site like bloom library, then we should get the blur and focus events and the data is good. If the host is something else (i.e., BR) and is coded to handle progress reports, it can easily be coded to ignore readDuration. (BTW, do we need to do something about readDuration in the Windows desktop reader? ReadingAppBuilder?) Is it just because ReadingAppBuilder hasn't (yet) been coded to handle readDuration? If that's the only known special case maybe it would be better to code explicitly for the cases where we DO want to delete it?
The comments in ReadMe.md reminded me that host might also be something like "embed@xyzzy.org". In that case IIRC a whole bloomlibrary website is embedded in an iframe...but we probably want readDuration from BP to be reported by the embedded BL?
Anyway, it just feels wrong that BP is coded to know specifically about bloomlibrary like this. 
Can we detect that we're running at the root of an Android browser from !window.parent? If window.parent is falsy, then either (a) we're on Android, and want to delete this, or (b) independent is true, and we're going to report for ourselves, and won't take this code branch at all, or (c) there's nothing to receive this message, so it doesn't matter what's in it. If window.parent is truthy and independent is false, we're in an iframe, so probably running in a bloom-library or similar iframe and getting blur/focus messages.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/230)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/externalContext.ts, line 92 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/230#-MglJB_0E9OsqsnNi-pL:-MglRICpAtV4UmOLQ_U5:bn0exrd) ([raw file](https://github.com/bloombooks/bloom-player/blob/ea6da1779114cfcb70898a6a91bf9f74fdd44422/src/externalContext.ts#L92)):*

I don't see why we ever delete it. If the host is a web site like bloom library, then we should get the blur and focus events and the data is good. If the host is something else (i.e., BR) and is coded to handle progress reports, it can easily be coded to ignore readDuration. (BTW, do we need to do something about readDuration in the Windows desktop reader? ReadingAppBuilder?) Is it just because ReadingAppBuilder hasn't (yet) been coded to handle readDuration? If that's the only known special case maybe it would be better to code explicitly for the cases where we DO want to delete it?
The comments in ReadMe.md reminded me that host might also be something like "embed@xyzzy.org". In that case IIRC a whole bloomlibrary website is embedded in an iframe...but we probably want readDuration from BP to be reported by the embedded BL?
Anyway, it just feels wrong that BP is coded to know specifically about bloomlibrary like this. 
Can we detect that we're running at the root of an Android browser from !window.parent? If window.parent is falsy, then either (a) we're on Android, and want to delete this, or (b) independent is true, and we're going to report for ourselves, and won't take this code branch at all, or (c) there's nothing to receive this message, so it doesn't matter what's in it. If window.parent is truthy and independent is false, we're in an iframe, so probably running in a bloom-library or similar iframe and getting blur/focus messages.
</blockquote></details>

In brief...
My thinking was that it is much safer to not report than to report incorrect/invalid numbers. And the problem with a blacklist is we don't know what might host this in the future and whether the host has been coded to ignore the incorrect/invalid numbers.

We could pass in another parameter to tell b-p whether we want readDuration (or something slightly more generic like if it will properly process focus/blur), but I'm not convinced that would be better since it seems to add unnecessary complexity.

For now, I'm happy to just leave RAB and BloomPUB Viewer not reporting, but maybe that is just being lazy.

Let's discuss when we're both available.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/230)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/externalContext.ts, line 92 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/230#-MglJB_0E9OsqsnNi-pL:-MgmEuMA7Q8RuTqINIkR:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/ea6da1779114cfcb70898a6a91bf9f74fdd44422/src/externalContext.ts#L92)):*

We decided the implementation I have is ok. I'm going to expand the comment some.
</blockquote></details>

Done.

PR #233, opened 2021-08-16
Workaround for scrollbar sometimes missing (BL-10217)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/233)

### DIFF
--- a/src/bloom-player.less
+++ b/src/bloom-player.less
@@ -196,6 +196,12 @@ We just make them follow the main content normally. */
     .bloom-editable.bloom-visibility-code-on {
         // Hidden because we let jquery.niceScroll control the scrollbars instead.
         overflow-y: hidden !important; // unfortunately this needs !important to override 2 basePage.less rules
+
+        // Flexboxes default to flex-shrink: 1, but we don't want paragraphs (or probably anything) to shrink if they're overflowing.
+        // (we add a scrollbar instead).  See BL-10217)
+        p {
+            flex-shrink: 0;
+        }
     }
     // it would be more elegant to embed this rule in the one above, but I can't figure out
     // how to write it so that the .bloom-textOverPicture has to be OUTSIDE the .bloomeditable


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/233)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[src/bloom-player-core.tsx, line 2061 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/233#-MhErqGk1nZYdlimkiPp:-MhErqGoA3VJTT-tm2_3:bvbbmb1) ([raw file](https://github.com/bloombooks/bloom-player/blob/0230e15e7153384cb9dc8737854dee19132aebf0/src/bloom-player-core.tsx#L2061)):*
> ```XML
>                 const marginOfErrorInPx = 0.5;
> 
>                 return overflowY > 0 - marginOfErrorInPx;
> ```

The error we were trying to fix is much more than half a pixel. Turns  out the cause is that the parent is display:flex and we're allowing it to shrink the paragraphs. I think correcting that (flex-shrink:0) might be a better fix.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/233)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/bloom-player-core.tsx, line 2061 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/233#-MhErqGk1nZYdlimkiPp:-MhEzwce8vGZFQzNvbML:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/0230e15e7153384cb9dc8737854dee19132aebf0/src/bloom-player-core.tsx#L2061)):*

The error we were trying to fix is much more than half a pixel. Turns  out the cause is that the parent is display:flex and we're allowing it to shrink the paragraphs. I think correcting that (flex-shrink:0) might be a better fix.
</blockquote></details>

Done.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/233)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #250, opened 2022-02-04
Allow badly-overflowing bubbles to scroll (BL-10536)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/250)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -2156,7 +2156,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
             let countOfObserversThatHaveReported = 0;
             $(bloomPage)
                 .find(
-                    ":not(.bloom-textOverPicture) > .bloom-translationGroup .bloom-editable.bloom-visibility-code-on"
+                    ".bloom-translationGroup .bloom-editable.bloom-visibility-code-on"
                 )
                 .each((index, elt) => {
                     // Process the blocks that are possibly overflowing.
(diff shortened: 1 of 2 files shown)


andrew-polk (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/250)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player.less, line 38 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/250#-MvL_GvDBqzwobL6Aamu:-MvL_GvDBqzwobL6Aamv:boxktia) ([raw file](https://github.com/bloombooks/bloom-player/blob/bbc0304533f11d86146c83ec4d1fec4d9f44d5aa/src/bloom-player.less#L38)):*
> ```Less
> .scrolling-bubble {
>     display: block !important;
> }
> ```

I guess we're pretty sure that we only need flex for vertical centering?
I.e. this isn't going to break something else?
___
*[src/bloom-player-core.tsx, line 2216 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/250#-MvLZzl91F0eT2hxPb1S:-MvLZzl91F0eT2hxPb1T:b-wbcsf9) ([raw file](https://github.com/bloombooks/bloom-player/blob/bbc0304533f11d86146c83ec4d1fec4d9f44d5aa/src/bloom-player-core.tsx#L2216)):*
> ```XML
>                                     // which makes the 'overflow' 13. A ratio of 0.5 as we originally
>                                     // proposed would give us a scroll bar we don't want.
>                                     let maxBubbleOverflowRatio = 0.6; // of a line
> ```

It is possible I'm just not understanding something, but this doesn't seem like a ratio to me.
Isn't it just a number? i.e. 0.6 of the height of a line?
___
*[src/bloom-player-core.tsx, line 2236 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/250#-MvLZsOdCTtwot0qV8ME:-MvLZsOdCTtwot0qV8MF:b31veak) ([raw file](https://github.com/bloombooks/bloom-player/blob/bbc0304533f11d86146c83ec4d1fec4d9f44d5aa/src/bloom-player-core.tsx#L2236)):*
> ```XML
>                                     const lineHeightPx = window.getComputedStyle(
>                                         entry.target
>                                     ).lineHeight;
> ```

I guess it is safe to assume we always get lineHeight in px?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/250)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[src/bloom-player.less, line 38 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/250#-MvL_GvDBqzwobL6Aamu:-MvPPHelEA2TGdcejwxY:bb3guzt) ([raw file](https://github.com/bloombooks/bloom-player/blob/bbc0304533f11d86146c83ec4d1fec4d9f44d5aa/src/bloom-player.less#L38)):*

I guess we're pretty sure that we only need flex for vertical centering?
I.e. this isn't going to break something else?
</blockquote></details>

I can't absolutely guarantee that. But display:flex, at least with vertical centering, definitely breaks nicescroll. I can try other ways to fix things, such as turning off whatever flex setting distributes the space top and bottom, if you think it's worth it.
___
*[src/bloom-player-core.tsx, line 2216 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/250#-MvLZzl91F0eT2hxPb1S:-MvPPtYl1RJb-M5hN1Sy:b-p4pv9o) ([raw file](https://github.com/bloombooks/bloom-player/blob/bbc0304533f11d86146c83ec4d1fec4d9f44d5aa/src/bloom-player-core.tsx#L2216)):*

It is possible I'm just not understanding something, but this doesn't seem like a ratio to me.
Isn't it just a number? i.e. 0.6 of the height of a line?
</blockquote></details>

I think of it as the ratio (maybe fraction would be better?) allowed overflow / line height. See if the new name works better.
___
*[src/bloom-player-core.tsx, line 2236 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/250#-MvLZsOdCTtwot0qV8ME:-MvPPaH09pR7z3wyjWt0:b-p2u7p3) ([raw file](https://github.com/bloombooks/bloom-player/blob/bbc0304533f11d86146c83ec4d1fec4d9f44d5aa/src/bloom-player-core.tsx#L2236)):*

I guess it is safe to assume we always get lineHeight in px?
</blockquote></details>

I believe so. We're doing this in various other places. It definitely doesn't depend on what unit the CSS used to set it. I suppose it MIGHT be browser-dependent, but I don't think it is.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/250)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #255, opened 2022-03-07
Support Bloom Editor recording book as video (BL-10991)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)

### DIFF
--- a/src/bloom-player-controls.tsx
+++ b/src/bloom-player-controls.tsx
@@ -12,7 +12,7 @@ import {
     setExternalControlCallback,
     logError
 } from "./externalContext";
-import { ControlBar, IExtraButton } from "./controlBar";
+import { ControlBar, IExtraButton, IVideoSettings } from "./controlBar";
 import { ThemeProvider } from "@material-ui/styles";
 import theme from "./bloomPlayerTheme";
 import React, { useState, useEffect, useRef, LegacyRef } from "react";
@@ -35,11 +35,13 @@ import { MuiThemeProvider } from "@material-ui/core/styles";
 // the pages is already known as a slider, so the two would get confused.
 import DragBar from "@material-ui/core/Slider";
 import { bloomRed } from "./bloomPlayerTheme";
+import { setDurationOfPagesWithoutNarration } from "./narration";
 
 // This component is designed to wrap a BloomPlayer with some controls
 // for things like pausing audio and motion, hiding and showing
 // image descriptions.
 
+export type autoPlayType = "yes" | "no" | "motion"; // default "motion" means autoplay only motion books
 interface IProps {
     // Url of the bloom book (folder). Should be a valid, well-formed URL
     // e.g. any special chars in the path or book title should be appropriately encoded using URL (percent) encoding
@@ -54,6 +56,16 @@ interface IProps {
     showContextPages?: boolean;
     // when bloom-player is told what content language to use from the start (vs. user changing using the language picker)
     initialLanguageCode?: string;
+    // when used in the Bloom Editor video recording publish panel, initializes the state of various
+    // settings. Not intended to be used with initialLanguageCode, as each can set the same property.
+    // Currently, this is a json string which can control whether to narrate image descriptions and
+    // also provides an alternative way to specify initialLanguageCode. That's a slightly awkward
+    // redundancy, but the intent is that videoSettings should be treated as an opaque string, and never
+    // created except by Bloom Player. It's purpose is so that the videoSettings sent to Bloom Editor
+    // using the /bloom/api/publish/video/videoSettings by the Preview instance of BP in the audio/video
+    // publish tab can be passed to the Recording instance. It should be possible to properly handle
+    // any additional settings without changing BloomEditor.
+    videoSettings?: string;
     paused: boolean;
     useOriginalPageSize?: boolean;
     // in production, this is just "". But during testing, we need
@@ -61,6 +73,26 @@ interface IProps {
     // e.g. src/activity-starter/
     locationOfDistFolder: string;
     extraButtons?: IExtraButton[];
+    // Hide the next/previous page buttons (mainly useful in autoplay='yes' scenarios like
+    // audio/video preview and recording)
+    hideSwiperButtons?: boolean;
+    // Options here allow Bloom Player to advance automatically through the book (or just some books, or not at all).
+    // Audio will play on the first page (even if not advancing automatically) unless autoplay is set to "no"
+    // (or unless the browser does not allow it).
+    autoplay: autoPlayType;
+    // If true, pages marked as bloom-interactive will not be shown. This is also useful in
+    // non-interactive scenarios like autoplaying a book to record video.
+    skipActivities?: boolean;
+    // This indicates that we're in the Preview pane of the audio/video recording publish tab
+    // in Bloom Editor. In this mode,
+    // - the nav bar is always shown
+    // - controls in it that are not relevant to how the book will be auto-played for recording are hidden
+    // - when those relevant controls are activated, the results are transmitted to Bloom Editor
+    // using the /bloom/api/publish/video/videoSettings API.
+    videoPreviewMode?: boolean;
+    // Send a report to Bloom API about sounds that have been played (when we reach the end
+    // of the book in autoplay).
+    shouldReportSoundLog?: boolean;
 }
 
 // This logic is not straightforward...
@@ -113,6 +145,7 @@ function getWindowSize() {
 
 export const BloomPlayerControls: React.FunctionComponent<IProps &
     React.HTMLProps<HTMLDivElement>> = props => {
+    const [autoplay, setAutoplay] = useState(props.autoplay);
     // default is to center BP vertically; various versions of blorg should pass this as false.
     const doVerticalCentering =
         props.centerVertically === undefined ? true : props.centerVertically;
@@ -128,6 +161,14 @@ export const BloomPlayerControls: React.FunctionComponent<IProps &
             setPaused(false);
         } else if (data.play) {
             setPaused(false);
+            if (data.autoplay) {
+                setAutoplay(data.autoplay);
+            }
+        } else if (data.reset) {
+            setPageNumberControlPos(0);
+            if (pageNumberSetter.current) {
+                pageNumberSetter.current(0);
+            }
         } else if (data.controlAction) {
             handleControlMessage(data.controlAction as string);
         }
@@ -571,17 +612,31 @@ export const BloomPlayerControls: React.FunctionComponent<IProps &
         setPageNumber: (pn: number) => void
     ): void => {
         pageNumberSetter.current = setPageNumber;
-        let languageCode: string;
+        let languageCode: string | undefined = undefined;
+        if (props.videoSettings) {
+            const videoSettings = JSON.parse(
+                props.videoSettings
+            ) as IVideoSettings;
+            languageCode = videoSettings.lang!;
+            setShouldReadImageDescriptions(
+                videoSettings.imageDescriptions ?? false
+            );
+        }
 
-        // This is the case where the url specified an initial language
-        if (
-            props.initialLanguageCode &&
-            bookLanguages.map(l => l.Code).includes(props.initialLanguageCode)
-        ) {
+        // This is the case where the url specified an initial language. Not normally done if
+        // videoSettings is passed, but will override it.
+        if (props.initialLanguageCode) {
             languageCode = props.initialLanguageCode;
+        }
+
+        if (
+            languageCode &&
+            bookLanguages.map(l => l.Code).includes(languageCode)
+        ) {
             LangData.selectNewLanguageCode(bookLanguages, languageCode);
         }
-        // this is the case where no initial language was specified in the url
+        // this is the case where no initial language was specified in the url or videoSettings
+        // (or, improbably, when they specified one that the book doesn't have)
         else {
             languageCode =
                 bookLanguages.length > 0 ? bookLanguages[0].Code : "";
@@ -589,6 +644,9 @@ export const BloomPlayerControls: React.FunctionComponent<IProps &
         setActiveLanguageCode(languageCode);
         setLanguageData(bookLanguages);
         setBookHasImageDescriptions(bookHasImageDescriptions);
+        if (props.videoPreviewMode && !showAppBar) {
+            setShowAppBar(true);
+        }
     };
 
     const playString = LocalizationManager.getTranslation(
@@ -717,6 +775,7 @@ export const BloomPlayerControls: React.FunctionComponent<IProps &
                 backClicked={() => onBackClicked()}
                 showPlayPause={hasAudio || hasMusic || hasVideo}
                 bookLanguages={languageData}
+                activeLanguageCode={activeLanguageCode}
                 onLanguageChanged={(isoCode: string) =>
                     handleLanguageChanged(isoCode)
                 }
@@ -728,6 +787,7 @@ export const BloomPlayerControls: React.FunctionComponent<IProps &
                     setShouldReadImageDescriptions(!shouldReadImageDescriptions)
                 }
                 nowReadingImageDescription={nowReadingImageDescription}
+                videoPreviewMode={props.videoPreviewMode}
             />
             <BloomPlayerCore
                 // We believe/hope we can do this a better way (without LegacyRef) once BloomPlayerCore is a function component.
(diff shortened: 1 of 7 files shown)


andrew-polk (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)* status: all files reviewed, 8 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player-controls.tsx, line 72 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxanQqJ3mvWA3IXUNN2:-MxanQqJ3mvWA3IXUNN3:b-w4m6jl) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L72)):*
> ```XML
>     autoplay: autoPlayType;
>     skipActivities?: boolean;
>     videoPreviewMode?: boolean;
> ```

Some of these settings may seem obvious by their names currently, but I think we will help ourselves to document them now.
Seems like the comment for `videoSettings` should also be expanded to include specifics on what settings can be set.

As a more general question, how did you decide which settings should be part of `videoSettings` and which should be separate?
For example, image descriptions could easily be a setting separate from video.
___
*[src/bloom-player-controls.tsx, line 75 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeOFur21c65pdaUJF7:-MxeOFur21c65pdaUJF8:b-uiru5g) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L75)):*
> ```XML
>     // Send a report to Bloom API about sounds that have been played (when we reach the end
>     // of the book in autoplay).
>     reportSoundLog?: boolean;
> ```

I think this might be better as `shouldReportSoundLog`. What do you think?

However, I see that we haven't been consistent at all with boolean names above. Many of them should probably start with `is`... so maybe not worth it?

But in this case, I got into a little confusion between `reportSoundLog` (boolean param) and `reportSoundsLogged` (function).



_Code quote:_
```XML
reportSoundLog
```
___
*[src/bloom-player-controls.tsx, line 772 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxamnLb-jyo3u6jgKXM:-MxamnLb-jyo3u6jgKXN:b-yjo863) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L772)):*
> ```XML
>                 videoPreviewMode={props.videoPreviewMode}
>             />
>             {props.videoPreviewMode}
> ```

Did this get left in by accident after debugging?



_Code quote:_
```XML
{props.videoPreviewMode}
```
___
*[src/controlBar.tsx, line 157 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeRuLx3ESUKrifO6UE:-MxeRuLx3ESUKrifO6UF:b-ec3rzf) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/controlBar.tsx#L157)):*
> ```XML
>                     "Content-Type": "text/plain"
>                 }
>             }
> ```

Is there a particular reason to stringify and send as text rather than sending as json?
___
*[src/externalContext.ts, line 238 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxePh6a7IZz0ek6DXZa:-MxePh6a7IZz0ek6DXZb:b-wprkh1) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/externalContext.ts#L238)):*
> ```XML
>     volume: Number;
>     startTime: string;
>     endTime?: string; // if not set, play the whole sound
> ```

Would this be better as 
`// if not set, the whole sound was played`
?



_Code quote:_
```XML
// if not set, play the whole sound
```
___
*[src/externalContext.ts, line 290 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxePve6BH4bXRDWhqx_:-MxePve6BH4bXRDWhqxa:b-d8gl0z) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/externalContext.ts#L290)):*
> ```XML
>     }
>     soundsPlayed = [];
> }
> ```

It feels odd for bloom-player to have knowledge of Bloom's api.
Did you weigh the pros/cons of just using a message?
Is there ever going to be another host which would want to have the sounds info? The only thing I can think of is if we ever wanted to make the video creator a stand-alone tool. But that doesn't seem likely at the moment.
___
*[src/externalContext.ts, line 290 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeQuUp4KWSPZ8I002D:-MxeQuUp4KWSPZ8I002E:b8i157q) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/externalContext.ts#L290)):*
> ```XML
>     }
>     soundsPlayed = [];
> }
> ```

As a second issue here, I would like to see more consistency with how we make calls to Bloom. Let's extract all the posts/messages to externalContext (or even a different file if we stay with posts?). Currently they are spread across three files.
Also, the call to start is conditional on the core prop. But here we call the function unconditionally and then check the param again inside the function.
___
*[src/music.ts, line 180 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeNs3r9IzZczNGUxM_:-MxeNs3r9IzZczNGUxMa:bka33l8) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/music.ts#L180)):*
> ```XML
>         this.getPlayer().currentTime = 0;
>         this.play();
>         this.musicLogIndex = logSoundRepeat(this.musicLogIndex);
> ```

Did you consider the option of just getting the src from the player? Or if that doesn't work, even keeping track of the currently playing src? I guess you have to track the volume level, too.
If that works, I think you can get rid of the index and even `logSoundRepeat` entirely.

Later: hm, I guess you need index for paused. Still, it might be worth it to get rid of logSoundRepeat.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)* status: 3 of 7 files reviewed, 8 unresolved discussions (waiting on @andrew-polk)

___
*[src/bloom-player-controls.tsx, line 72 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxanQqJ3mvWA3IXUNN2:-Mxe_8h73xfqebCYB4I-:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L72)):*

Some of these settings may seem obvious by their names currently, but I think we will help ourselves to document them now.
Seems like the comment for `videoSettings` should also be expanded to include specifics on what settings can be set.

As a more general question, how did you decide which settings should be part of `videoSettings` and which should be separate?
For example, image descriptions could easily be a setting separate from video.
</blockquote></details>

Done.
___
*[src/bloom-player-controls.tsx, line 75 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeOFur21c65pdaUJF7:-Mxe_PSX6Wvi-nPlk4qe:bn0k8d3) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L75)):*

I think this might be better as `shouldReportSoundLog`. What do you think?

However, I see that we haven't been consistent at all with boolean names above. Many of them should probably start with `is`... so maybe not worth it?

But in this case, I got into a little confusion between `reportSoundLog` (boolean param) and `reportSoundsLogged` (function).
</blockquote></details>

I think I agree, though currently I have not also renamed the URL param. It's awkward because many of the options could be named like this (shouldAutoplay, shouldHideSwiperButtons...) but the extra length doesn't quite seem justified for most of them (to me, anyway). In the URL, however, the meaning of reportSoundLog is not ambiguous and the URL is already cumbersome enough and I think it would be more confusing and make it less consistent with the other param names.
___
*[src/bloom-player-controls.tsx, line 772 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxamnLb-jyo3u6jgKXM:-MxeazbgAhO0jDJkPgys:b-uyzobq) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L772)):*

Did this get left in by accident after debugging?
</blockquote></details>

I think it must have.
___
*[src/controlBar.tsx, line 157 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeRuLx3ESUKrifO6UE:-MxecJphDetL9AH5U_v_:bcng87f) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/controlBar.tsx#L157)):*

Is there a particular reason to stringify and send as text rather than sending as json?
</blockquote></details>

I think so and have tried to explain in a comment. John particularly wanted to make as much use as we can of the playback settings that BP already has controls for, without having to make BE aware of each of them.
___
*[src/externalContext.ts, line 238 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxePh6a7IZz0ek6DXZa:-Mxed9QH1jA5ritRFYq8:b-7um1et) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/externalContext.ts#L238)):*

Would this be better as 
`// if not set, the whole sound was played`
?
</blockquote></details>

Kind of depends on whether your point of view is the sender or the receiver. I've tried to rewrite it to cover both.
___
*[src/externalContext.ts, line 290 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxePve6BH4bXRDWhqx_:-MxedKX-Ai_GcPH2kFRw:b-55x4bo) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/externalContext.ts#L290)):*

It feels odd for bloom-player to have knowledge of Bloom's api.
Did you weigh the pros/cons of just using a message?
Is there ever going to be another host which would want to have the sounds info? The only thing I can think of is if we ever wanted to make the video creator a stand-alone tool. But that doesn't seem likely at the moment.
</blockquote></details>

I thought quite hard about it. I don't know of any way BE can receive messages that BP posts with postMessage. So we'd need another whole layer of wrapper (in which BP is an iframe) just to listen for the messages and pass them on to BE through the API. If we want the code in that wrapper to be typescript, it's going to end up loading the whole common bundle, and then we have to figure how to get the right URL for the player through the wrapper to the iframe...it could be done but felt like YAGNI.
Another problem is that in the case of telling Bloom when to start ffmpeg, BP needs to know that task is complete; but postMessage doesn't support waiting for a response, so we'd have to configure another message...
___
*[src/externalContext.ts, line 290 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeQuUp4KWSPZ8I002D:-MxejO7l2ka89g0t8Toi:bkxk4j6) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/externalContext.ts#L290)):*

As a second issue here, I would like to see more consistency with how we make calls to Bloom. Let's extract all the posts/messages to externalContext (or even a different file if we stay with posts?). Currently they are spread across three files.
Also, the call to start is conditional on the core prop. But here we call the function unconditionally and then check the param again inside the function.
</blockquote></details>

I think I've done what you're suggesting.
___
*[src/music.ts, line 180 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeNs3r9IzZczNGUxM_:-MxekCUD-lVG9Q3rvDfM:b8l7x9t) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/music.ts#L180)):*

Did you consider the option of just getting the src from the player? Or if that doesn't work, even keeping track of the currently playing src? I guess you have to track the volume level, too.
If that works, I think you can get rid of the index and even `logSoundRepeat` entirely.

Later: hm, I guess you need index for paused. Still, it might be worth it to get rid of logSoundRepeat.
</blockquote></details>

I don't see a good way to not need the index to make pause work. Given that, I don't see a lot of advantage in getting rid of logSoundRepeat. (Though I'm starting to wonder if we should have a new SoundLog.ts. These functions don't fit well in any of the current files. What do you think?)

andrew-polk (changes requested):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @JohnThomson)

___
*[src/music.ts, line 180 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeNs3r9IzZczNGUxM_:-MxexRK74ZVUDUgmF_nZ:b-e00fzu) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/music.ts#L180)):*

I don't see a good way to not need the index to make pause work. Given that, I don't see a lot of advantage in getting rid of logSoundRepeat. (Though I'm starting to wonder if we should have a new SoundLog.ts. These functions don't fit well in any of the current files. What do you think?)
</blockquote></details>

Yes, I think it would be worth extracting stuff to SoundLog.ts.
I would strongly consider moving the bloom api stuff there, too. On the one hand, it isn't necessarily related to sound logs in the future, but on the other hand, it currently is. And it would be nice to remove it from the more general externalContext context.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)* status: all files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[src/bloom-player-controls.tsx, line 72 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxanQqJ3mvWA3IXUNN2:-Mxf9VbD0GqaR-BlCkX5:bgos95o) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L72)):*

Done.
</blockquote></details>

I think you missed the second part of my comment (the question about settings... and should imagedesc be its own setting altogether)

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)* status: 2 of 8 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[src/bloom-player-controls.tsx, line 72 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxanQqJ3mvWA3IXUNN2:-MxfB8YM25vXfrJ0c5sY:b-1qh4do) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/bloom-player-controls.tsx#L72)):*

I think you missed the second part of my comment (the question about settings... and should imagedesc be its own setting altogether)
</blockquote></details>

I wrote a long comment on the videoSettings which attempts to answer that. It's true that whether to play image descriptions could plausibly be a separate setting, but I'm applying yagni. Meanwhile, the things in videoSettings are the things you can control with BP control bar that are relevant to video recording playback.
___
*[src/music.ts, line 180 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/255#-MxeNs3r9IzZczNGUxM_:-Mxfbwin1CTIgT5ia266:b-4u3nkb) ([raw file](https://github.com/bloombooks/bloom-player/blob/f466292149aa3620957ab8a42de4fce5436b6de1/src/music.ts#L180)):*

Yes, I think it would be worth extracting stuff to SoundLog.ts.
I would strongly consider moving the bloom api stuff there, too. On the one hand, it isn't necessarily related to sound logs in the future, but on the other hand, it currently is. And it would be nice to remove it from the more general externalContext context.
</blockquote></details>

The messages are not all sound-related, so I called the new file videoRecordingSupport

andrew-polk (approved):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

andrew-polk (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/255)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #264, opened 2022-04-08
Make the animation continue into page fade (BL-11100)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/264)

### DIFF
--- a/src/animation.ts
+++ b/src/animation.ts
@@ -433,7 +433,13 @@ export class Animation {
                     "; animation-name: " +
                     movePicName +
                     "; animation-duration: " +
-                    this.animationDuration +
+                    // Making the animation take an extra second lets the movement
+                    // continue smoothly into the fade to the next page.
+                    // (Currently all these these things go together: if we are doing animation,
+                    // then we are autoplaying and we fade from one page to the next. If these things
+                    // become separately controllable, we might not always want an extra second
+                    // of animation.)
+                    (this.animationDuration + 1) +
                     "s; animation-fill-mode: forwards; " +
                     "animation-timing-function: linear;}",
                 1


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/264)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/animation.ts, line 438 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/264#-N-9wWFu2VITenjC0OKc:-N-9wWFu2VITenjC0OKd:boiycdv) ([raw file](https://github.com/bloombooks/bloom-player/blob/9de9ce8930a24362b0154b39395b52f34f549dea/src/animation.ts#L438)):*
> ```XML
>                     // Making the animation take an extra second lets the movement
>                     // continue smoothly into the fade to the next page.
>                     (this.animationDuration + 1) +
> ```

We only want this when fading, right.
I think in practice we never do animation without autoplay (which will now always fade).
So I think with current code we are ok. 
I just find myself wondering if we should be more explicit to say this is tied to autoplay (in case the code changes in the future). But I'm open to it being yagni.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/264)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/animation.ts, line 438 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/264#-N-9wWFu2VITenjC0OKc:-N-NpQ7O4d-Hs_HObXYJ:b-896fix) ([raw file](https://github.com/bloombooks/bloom-player/blob/9de9ce8930a24362b0154b39395b52f34f549dea/src/animation.ts#L438)):*

We only want this when fading, right.
I think in practice we never do animation without autoplay (which will now always fade).
So I think with current code we are ok. 
I just find myself wondering if we should be more explicit to say this is tied to autoplay (in case the code changes in the future). But I'm open to it being yagni.
</blockquote></details>

Done.

PR #281, opened 2022-09-13
Disable highlight of multiple spaces (BL-11485)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/281)

### DIFF
--- a/src/bloom-player.less
+++ b/src/bloom-player.less
@@ -47,7 +47,7 @@ above the page so that page numbers will increment.*/
 // Actually: this did not prevent it from applying in Firefox,: @media screen and (-webkit-min-device-pixel-ratio: 0) {
 // Nor did this: @supports (-webkit-appearance: none) {
 /*@media screen and (-webkit-min-device-pixel-ratio: 0) {
-    span.ui-audioCurrent:not(.disableHighlight) {
+    span.ui-audioCurrent:not(.ui-suppressHighlight) {
         background: linear-gradient(@audioHighlighting, @audioHighlighting);
         background-blend-mode: overlay;
         mix-blend-mode: overlay;
@@ -57,11 +57,15 @@ above the page so that page numbers will increment.*/
 // The above doesn't work on Firefox (through FF 68) (or in Bloom or 2019-era Edge). For those, we can at least
 // deal with the most common case of overlap (the book title on the cover) by just
 // treating the text as a rectangle. And it's fine that this will show this way in Chrome too:
-.Title-On-Cover-style span.ui-audioCurrent:not(.disableHighlight) {
+.Title-On-Cover-style span.ui-audioCurrent:not(.ui-suppressHighlight) {
     display: inline-block;
 }
-span.ui-audioCurrent:not(.disableHighlight),
-div.ui-audioCurrent:not(.disableHighlight) p {
+span.ui-audioCurrent:not(.ui-suppressHighlight, .ui-disableHighlight),
+div.ui-audioCurrent:not(.ui-suppressHighlight, .ui-disableHighlight) p {
+    background-color: @audioHighlighting;
+}
+
+.ui-audioCurrent .ui-enableHighlight {
     background-color: @audioHighlighting;
 }
 
--- a/src/bloom-player.less
+++ b/src/bloom-player.less
@@ -47,7 +47,7 @@ above the page so that page numbers will increment.*/
 // Actually: this did not prevent it from applying in Firefox,: @media screen and (-webkit-min-device-pixel-ratio: 0) {
 // Nor did this: @supports (-webkit-appearance: none) {
 /*@media screen and (-webkit-min-device-pixel-ratio: 0) {
-    span.ui-audioCurrent:not(.ui-suppressHighlight) {
+    span.ui-audioCurrent:not(.ui-suppressHighlight):not(.ui-disableHighlight) {
         background: linear-gradient(@audioHighlighting, @audioHighlighting);
         background-blend-mode: overlay;
         mix-blend-mode: overlay;
@@ -57,11 +57,14 @@ above the page so that page numbers will increment.*/
 // The above doesn't work on Firefox (through FF 68) (or in Bloom or 2019-era Edge). For those, we can at least
 // deal with the most common case of overlap (the book title on the cover) by just
 // treating the text as a rectangle. And it's fine that this will show this way in Chrome too:
-.Title-On-Cover-style span.ui-audioCurrent:not(.ui-suppressHighlight) {
+.Title-On-Cover-style
+    span.ui-audioCurrent:not(.ui-suppressHighlight):not(.ui-disableHighlight) {
     display: inline-block;
 }
-span.ui-audioCurrent:not(.ui-suppressHighlight, .ui-disableHighlight),
-div.ui-audioCurrent:not(.ui-suppressHighlight, .ui-disableHighlight) p {
+// Note: :not(classA, classB) syntax not supported by some browsers, notably FF60.
+// Use :not(classB):not(classB) syntax instead!!!
+span.ui-audioCurrent:not(.ui-suppressHighlight):not(.ui-disableHighlight),
+div.ui-audioCurrent:not(.ui-suppressHighlight):not(.ui-disableHighlight) p {
     background-color: @audioHighlighting;
 }
 
--- a/src/narration.ts
+++ b/src/narration.ts
@@ -5,6 +5,20 @@ import { SwiperInstance } from "react-id-swiper";
 import { logSound } from "./videoRecordingSupport";
 
 const kSegmentClass = "bloom-highlightSegment";
+
+// Indicates that the element should be highlighted.
+const kEnableHighlightClass = "ui-enableHighlight";
+
+// Indicates that the element should NOT be highlighted.
+// For example, some elements have highlighting prevented at this level
+// because its content has been broken into child elements, only some of which show the highlight
+const kDisableHighlightClass = "ui-disableHighlight";
+
+// Indicates that highlighting is briefly/temporarily suppressed,
+// but may become highlighted later.
+// For example, audio highlighting is suppressed until the related audio starts playing (to avoid flashes)
+const kSuppressHighlightClass = "ui-suppressHighlight";
+
 var durationOfPagesWithoutNarration = 3.0; // seconds
 export function setDurationOfPagesWithoutNarration(d: number) {
     durationOfPagesWithoutNarration = d;
@@ -89,6 +103,8 @@ export default class Narration {
         // (Deals with the case where you are on a page with audio, switch to a page without audio, then switch back to original page)
         ++this.currentAudioSessionNum;
 
+        this.fixHighlighting();
+
         // Sorted into the order we want to play them, then reversed so we
         // can more conveniently pop the next one to play from the end of the stack.
         this.elementsToPlayConsecutivelyStack = sortAudioElements(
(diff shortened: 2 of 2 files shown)


sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/281)* status: 0 of 2 files reviewed, 2 unresolved discussions

___
*[`src/narration.ts` line 13 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBr_oODCxaEbJRCJuya:-NBr_oODCxaEbJRCJuyb:btaqrou) ([raw file](https://github.com/bloombooks/bloom-player/blob/68e9c34cd1997fee7ba71d958361d16d4c6a6215/src/narration.ts#L13)):*
> ```XML
> 
> // Indicates that the element should NOT be highlighted.
> const kDisableHighlightClass = "ui-disableHighlight";
> ```

I'm not super confident about how to decide which class prefixes get used ".bloom-[...] " vs ".ui-[...]" vs "[noPrefix]" but... I think it wasn't right that we used no prefix before? I think it should have a ".ui-" prefix?
___
*[`src/narration.ts` line 17 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBr_pgf-eEXwiLRn99b:-NBr_pgf-eEXwiLRn99c:b-fbu4iy) ([raw file](https://github.com/bloombooks/bloom-player/blob/68e9c34cd1997fee7ba71d958361d16d4c6a6215/src/narration.ts#L17)):*
> ```XML
> // Indicates that highlighting is briefly/temporarily suppressed,
> // but may become highlighted later
> const kSuppressHighlightClass = "ui-suppressHighlight";
> ```

This is what was formerly known as "disableHighlight". I had some difficulty because we had 2 different senses of "disabling" things. (It was easier to have them as 2 different classes to make less interference in the logic/timing between them). The best solution to me seemed to be renaming the existing one, unfortunately.

JohnThomson (changes requested):
Nice piece of work. A few minor suggestions, and one that I think is more important.
And one more: this feels like the sort of thing that deserves unit testing. I'm not sure we've even started that for BP, but maybe this is a good time?

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/281#-:-NByECJfCDyz0A_WJwJt:b5q0qau)* status: all files reviewed, 5 unresolved discussions (waiting on @sujeffreyl)

___
*[`src/bloom-player.less` line 50 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy-dlXBrNHhPRnarYI:-NBy-dlXBrNHhPRnarYJ:b-a7la9o) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/bloom-player.less#L50)):*
> ```Less
> // Nor did this: @supports (-webkit-appearance: none) {
> /*@media screen and (-webkit-min-device-pixel-ratio: 0) {
>     span.ui-audioCurrent:not(.ui-suppressHighlight) {
> ```

Why do we need both :not's below but not here (or in TitleOnCover below)?
___
*[`src/narration.ts` line 13 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBr_oODCxaEbJRCJuya:-NBy-zj_8bAqAZicdU_1:b-vauzus) ([raw file](https://github.com/bloombooks/bloom-player/blob/68e9c34cd1997fee7ba71d958361d16d4c6a6215/src/narration.ts#L13)):*

I'm not super confident about how to decide which class prefixes get used ".bloom-[...] " vs ".ui-[...]" vs "[noPrefix]" but... I think it wasn't right that we used no prefix before? I think it should have a ".ui-" prefix?
</blockquote></details>

I think the idea is that bloom- classes are meaningful to bloom (used in bloom code, require that cs and ts code be checked before changing them) _and persistent_ in the document. ui- would generally also convey this, along with the idea that they don't need to be a permanent part of the document, but are only used while the page is being edited.
So, audio-sentence should have been bloom-audio-sentence (but too hard to change), but I think ui-suppressHighlight is right, and so is ui-disableHighlight, since both are just temporary.
Actually, bloom-player doesn't really need the convention (which may be why we didn't do it for disableHighlight) since changes it makes to the DOM are never saved anywhere. So I'd be OK with or without those prefixes. However, there's some benefit in knowing that a class is added by the player, not something it found in the source file, so I slightly prefer what you've done.
___
*[`src/narration.ts` line 17 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBr_pgf-eEXwiLRn99b:-NByBEzeCWhD-7ImMhqH:b-a2rufs) ([raw file](https://github.com/bloombooks/bloom-player/blob/68e9c34cd1997fee7ba71d958361d16d4c6a6215/src/narration.ts#L17)):*

This is what was formerly known as "disableHighlight". I had some difficulty because we had 2 different senses of "disabling" things. (It was easier to have them as 2 different classes to make less interference in the logic/timing between them). The best solution to me seemed to be renaming the existing one, unfortunately.
</blockquote></details>

I'd be inclined to make the comments (maybe even the name?) even more explicit: "highlighting is suppressed until the related audio starts playing" vs "highlighting is prevented at this level because its content has been broken into child elements only some of which show the highlight". How about ui-delayHighlightForAudio and ui-deferHighlightToChildren (or is 'delegate' better than 'defer'?)? (Assuming we aren't likely to want to use either with some other meaning, of course.)
___
*[`src/narration.ts` line 223 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy5UzH2wJgaMGzOuVQ:-NBy5UzH2wJgaMGzOuVR:b-g76szs) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/narration.ts#L223)):*
> ```XML
>                         this.multiSpaceRegexGlobal.lastIndex -
>                         matchingText.length,
>                     endIndex: this.multiSpaceRegexGlobal.lastIndex // exclusive index, not inclusive
> ```

Not sure if we've ever done this in JS, but Flex had a naming convention that an exclusive end index was called a limit (usually abbreviated lim). (If you don't see this anywhere it's probably not worth starting now, though.)
___
*[`src/narration.ts` line 270 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy6zpC4gm4-ZR0c7vs:-NBy6zpC4gm4-ZR0c7vt:b59jivg) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/narration.ts#L270)):*
> ```XML
>                 );
>                 elementToInsertBefore = nodeToInsert;
>             }
> ```

This seems unnecessarily complex. Why not just insert them in the original order, each before textNode?
___
*[`src/narration.ts` line 280 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy7VhP3ToylQn-n9Vu:-NBy7VhP3ToylQn-n9Vv:b-dcayje) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/narration.ts#L280)):*
> ```XML
>             // all that would happen is the child would allow the parent's highlight color to show through,
>             // which doesn't achieve what we want :(
>             parentElement?.classList?.add(kDisableHighlightClass);
> ```

This doesn't seem like enough. Aren't there cases (e.g., where TextNode is the child of an "em" rather than directly in the span that has the ui-audioCurrent class, or where the  ui-audioCurrent class is up on an unsplit text-box div) where the element where audio-current needs disabling is not the immediate parent? (I think closest(". ui-audioCurrent") would find the place you really want to put the disable.)

sujeffreyl (commented):
Let me see what I can do about the unit tests.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/281#-:-NC0Y3_sAjdLIn02iTbW:b-zhwoom)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @JohnThomson and @sujeffreyl)

___
*[`src/bloom-player.less` line 50 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy-dlXBrNHhPRnarYI:-NC0PR5o01HDBSEP3rKY:b-886gsx) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/bloom-player.less#L50)):*

Why do we need both :not's below but not here (or in TitleOnCover below)?
</blockquote></details>

Fixed
___
*[`src/narration.ts` line 13 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBr_oODCxaEbJRCJuya:-NC0PacG0SBvS0pRoF1I:b686jxx) ([raw file](https://github.com/bloombooks/bloom-player/blob/68e9c34cd1997fee7ba71d958361d16d4c6a6215/src/narration.ts#L13)):*

I think the idea is that bloom- classes are meaningful to bloom (used in bloom code, require that cs and ts code be checked before changing them) _and persistent_ in the document. ui- would generally also convey this, along with the idea that they don't need to be a permanent part of the document, but are only used while the page is being edited.
So, audio-sentence should have been bloom-audio-sentence (but too hard to change), but I think ui-suppressHighlight is right, and so is ui-disableHighlight, since both are just temporary.
Actually, bloom-player doesn't really need the convention (which may be why we didn't do it for disableHighlight) since changes it makes to the DOM are never saved anywhere. So I'd be OK with or without those prefixes. However, there's some benefit in knowing that a class is added by the player, not something it found in the source file, so I slightly prefer what you've done.
</blockquote></details>

Ok, thanks. Well, I'd like it to be easily portable to BloomDesktop, so I'll keep it with ui-*
___
*[`src/narration.ts` line 223 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy5UzH2wJgaMGzOuVQ:-NC0XfMzBbcfXgY4kLob:b4wqho9) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/narration.ts#L223)):*

Not sure if we've ever done this in JS, but Flex had a naming convention that an exclusive end index was called a limit (usually abbreviated lim). (If you don't see this anywhere it's probably not worth starting now, though.)
</blockquote></details>

Hmm, I looked around, but don't see that convention in our code, but I see from the Javascript String docs, it refers to this idea is a "indexEnd: The index of the first character to exclude from [...]". I updated my comment to match that

sujeffreyl (commented):
So I tried this, but... well, constructing a narration() object seems to lead to some trouble. I think it's related to ES modules. I tried to follow the instructions for enabling ES modules but well, I couldn't get it to work yet.
I figure we can try adding the tests on the BloomDesktop end instead.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/281#-:-NC0aoGHAQwW3S2LRZYt:b828gud)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @JohnThomson and @sujeffreyl)

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/281)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/narration.ts` line 17 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBr_pgf-eEXwiLRn99b:-NC1Pufj4ew45yH2yr1s:b-w30ifn) ([raw file](https://github.com/bloombooks/bloom-player/blob/68e9c34cd1997fee7ba71d958361d16d4c6a6215/src/narration.ts#L17)):*

I'd be inclined to make the comments (maybe even the name?) even more explicit: "highlighting is suppressed until the related audio starts playing" vs "highlighting is prevented at this level because its content has been broken into child elements only some of which show the highlight". How about ui-delayHighlightForAudio and ui-deferHighlightToChildren (or is 'delegate' better than 'defer'?)? (Assuming we aren't likely to want to use either with some other meaning, of course.)
</blockquote></details>

Ok, i made the comments more specific.

For the class names, I'm a fan of them more general purpose.
I kinda like "delay", but decided to keep suppress, more general purpose.
___
*[`src/narration.ts` line 270 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy6zpC4gm4-ZR0c7vs:-NC0eJDrCL8wJGlyCDgN:b-lvlsho) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/narration.ts#L270)):*

This seems unnecessarily complex. Why not just insert them in the original order, each before textNode?
</blockquote></details>

Ah, yes that works. Thanks!
___
*[`src/narration.ts` line 280 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy7VhP3ToylQn-n9Vu:-NC0oNneARoNF0UWTERt:b-cu1rl6) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/narration.ts#L280)):*

This doesn't seem like enough. Aren't there cases (e.g., where TextNode is the child of an "em" rather than directly in the span that has the ui-audioCurrent class, or where the  ui-audioCurrent class is up on an unsplit text-box div) where the element where audio-current needs disabling is not the immediate parent? (I think closest(". ui-audioCurrent") would find the place you really want to put the disable.)
</blockquote></details>

It actually does seem like it's enough, but oftentimes by luck rather than design.
"em" case (partial text italicized): The em receives the disable-highlight (unnecessarily) in this case, but the non-italicized text sets disable-highlight on the correct element, yay.
"em" case (all text italicized): The em wraps the span instead of the other way around. Dodged a bullet there.
Unsplit text box: By design. This case doesn't matter or even match the earlier selectors. The entire paragraph gets highlighted, this is a block-level element highlighting and its highlighting borders will extend wider than the text regardless of multiple whitespace or not.

I do agree with your point though that the closest .ui-audioCurrent is more representative of what we're actually after than the parentElement though, so I updated that.

(Although, it's not so simple, because this is before .ui-audioCurrent is applied).

JohnThomson (approved):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/281)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

___
*[`src/narration.ts` line 280 at r2](https://reviewable.io/reviews/bloombooks/bloom-player/281#-NBy7VhP3ToylQn-n9Vu:-NC75YNv7zxoso3ViRmh:bcnx6iu) ([raw file](https://github.com/bloombooks/bloom-player/blob/757d622530cb594447058bee2916c3965cdecd30/src/narration.ts#L280)):*

It actually does seem like it's enough, but oftentimes by luck rather than design.
"em" case (partial text italicized): The em receives the disable-highlight (unnecessarily) in this case, but the non-italicized text sets disable-highlight on the correct element, yay.
"em" case (all text italicized): The em wraps the span instead of the other way around. Dodged a bullet there.
Unsplit text box: By design. This case doesn't matter or even match the earlier selectors. The entire paragraph gets highlighted, this is a block-level element highlighting and its highlighting borders will extend wider than the text regardless of multiple whitespace or not.

I do agree with your point though that the closest .ui-audioCurrent is more representative of what we're actually after than the parentElement though, so I updated that.

(Although, it's not so simple, because this is before .ui-audioCurrent is applied).
</blockquote></details>

I could imagine wanting the large blocks of white space to be un-highlighted even in an unsplit text-box. But, this is a temporary work-around; for now, if they want this enhancement, they'll have to split them. That's acceptable, it's not worth more of your time. So I'll go ahead and merge.

PR #283, opened 2022-09-21
Fix changing language on an activity page (BL-11449)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/283)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -620,7 +620,8 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                     });
             }
 
-            // If the user changes the image description button on the controlbar
+            // If the user changes the image description button on the controlbar or changes
+            // the active language for displaying the book
             // Review JohnT: I don't see why we want to do this simply on a transition from loading
             // to not-loading. We already do it on a timeout in the code that sets loading to false,
             // and we suppress effects on setIndex and showingPage until that timeout completes and
@@ -629,7 +630,9 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                 !this.state.isLoading &&
                 (prevState.isLoading ||
                     prevProps.shouldReadImageDescriptions !==
-                        this.props.shouldReadImageDescriptions)
+                        this.props.shouldReadImageDescriptions ||
+                    prevProps.activeLanguageCode !==
+                        this.props.activeLanguageCode)
             ) {
                 if (this.finishUpCalled) {
                     // We need to reset the page enough to get the narration rebuilt.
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
Probably fine, but I'm having a had time convincing myself.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/283#-:-NCtpb6o3G6TpEQ_c_DA:bjsuj6w)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/bloom-player-core.tsx` line 848 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/283#-NCtocMT1t3xpePtRTB7:-NCtocMT1t3xpePtRTB8:b-4bdk07) ([raw file](https://github.com/bloombooks/bloom-player/blob/07e83f19c9c728e76b0880a3abfc10c1280d7b41/src/bloom-player-core.tsx#L848)):*
> ```XML
>                     this.swiperInstance.slides[i] = page;
>                 }
>             }
> ```

I think this block needs more explanation (i.e., comments). Some questions that come to mind (given that I'm rusty on all this code)
- how does the page come to be stale (and in what sense)?
- why is this only relevant for activity pages?
- what goes wrong with the language switch if we don't do this (perhaps reference YouTrack issue)?

What will this do if the user has already answered a question of made progress with an activity in the original language?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/283)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/bloom-player-core.tsx` line 848 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/283#-NCtocMT1t3xpePtRTB7:-NCuiti5BOlaEYxDBSGb:bkq6xta) ([raw file](https://github.com/bloombooks/bloom-player/blob/07e83f19c9c728e76b0880a3abfc10c1280d7b41/src/bloom-player-core.tsx#L848)):*

I think this block needs more explanation (i.e., comments). Some questions that come to mind (given that I'm rusty on all this code)
- how does the page come to be stale (and in what sense)?
- why is this only relevant for activity pages?
- what goes wrong with the language switch if we don't do this (perhaps reference YouTrack issue)?

What will this do if the user has already answered a question of made progress with an activity in the original language?
</blockquote></details>

I added an extensive comment and removed one line of code that doesn't seem to be needed.
If the current inconsistent behavior of preserving/losing previous choices when changing language is undesirable, then presumably things can be made consistent one way or another.

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/283#-:-NCumnzVBh1sFV9x7ryw:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #286, opened 2022-10-03
Avoid horizontal scrolling to show almost-visible text (BL-11038)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/286)

### DIFF
--- a/src/narration.ts
+++ b/src/narration.ts
@@ -631,7 +631,30 @@ export default class Narration {
             return;
         }
 
-        element.scrollIntoView({
+        let mover = element as HTMLElement; // by default make the element itself scrollIntoView
+        if (
+            window.getComputedStyle(element.parentElement!).position !==
+            "static"
+        ) {
+            // We can make a new element absolutely positioned and it will be relative to the parent.
+            // The idea is to make an element much narrower than the element we are
+            // trying to make visible, since we don't want horizontal movement. Quite possibly,
+            // as in BL-11038, only some white space is actually off-screen. But even if the author
+            // has positioned a bubble so some text is cut off, we don't want horizontal scrolling,
+            // which inside swiper will weirdly pull in part of the next page.
+            // (In the pathological case that the bubble is more than half hidden, we'll do the
+            // horizontal scroll, despite the ugliness of possibly showing part of the next page.)
+            mover = document.createElement("div");
+            mover.style.position = "absolute";
+            mover.style.top = element.clientTop + "px";
+            mover.style.left =
+                element.clientLeft + element.clientWidth / 2 + "px";
+            mover.style.height = element.clientHeight + "px";
+            mover.style.width = "0";
+            element.parentElement?.insertBefore(mover, element);
+        }
+
+        mover.scrollIntoView({
             // Animated instead of sudden
             behavior: "smooth",
 
@@ -642,7 +665,11 @@ export default class Narration {
             block: "nearest"
 
             // horizontal alignment is controlled by "inline". We'll leave it as its default ("nearest")
+            // which typically won't move things at all horizontally
         });
+        if (mover !== element) {
+            mover.parentElement?.removeChild(mover);
+        }
     }
 
     // Returns true if swiping to this page is still in progress.


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/286)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/narration.ts` line 646 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/286#-NDYv-3C3QSI70NNkvBi:-NDYv-3D-a84trqdiK1y:b55acvf) ([raw file](https://github.com/bloombooks/bloom-player/blob/46c018749ca0461724fd0879e27280143dda06f9/src/narration.ts#L646)):*
> ```XML
>             // which inside swiper will weirdly pull in part of the next page.
>             // (In the pathological case that the bubble is more than half hidden, we'll do the
>             // horizontal scroll, despite the ugliness of possibly showing part of the next page.)
> ```

Can we instead say left = 0 and get rid of the special pathological case?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/286)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/narration.ts` line 646 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/286#-NDYv-3C3QSI70NNkvBi:-NDZRrAuBJX_DV7zB118:b-gtqyr9) ([raw file](https://github.com/bloombooks/bloom-player/blob/46c018749ca0461724fd0879e27280143dda06f9/src/narration.ts#L646)):*

Can we instead say left = 0 and get rid of the special pathological case?
</blockquote></details>

The bubble is just as likely to be partly off the left side of the page, in which case left=0 will cause it to scroll slightly that way.

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/286)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/narration.ts` line 646 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/286#-NDYv-3C3QSI70NNkvBi:-NDZYgdA-9YmMaQuDzn0:b-4h677w) ([raw file](https://github.com/bloombooks/bloom-player/blob/46c018749ca0461724fd0879e27280143dda06f9/src/narration.ts#L646)):*

The bubble is just as likely to be partly off the left side of the page, in which case left=0 will cause it to scroll slightly that way.
</blockquote></details>

Isn't the parent of the thing you're adding the page? So left:0 would be the first pixel of the page?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/286)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/narration.ts` line 646 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/286#-NDYv-3C3QSI70NNkvBi:-NDZptvbEK9SxWFLcO-2:bpxrp5d) ([raw file](https://github.com/bloombooks/bloom-player/blob/46c018749ca0461724fd0879e27280143dda06f9/src/narration.ts#L646)):*

Isn't the parent of the thing you're adding the page? So left:0 would be the first pixel of the page?
</blockquote></details>

No. We're typically trying to scrollIntoView a paragraph whose parent is a bloom-editable which (being in a bubble) may be partly off-screen in any direction. It may also be possible that we're trying to scrollIntoView a span whose parent is a paragraph inside a BE. So left:0 is the far left of the Bloom-editable (or maybe the paragraph), which (with centered text especially) may be a bit further left than the first visible character.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/286)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

___
*[`src/narration.ts` line 646 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/286#-NDYv-3C3QSI70NNkvBi:-NDZs-sr5RLXaX_ln74A:b-2fecm0) ([raw file](https://github.com/bloombooks/bloom-player/blob/46c018749ca0461724fd0879e27280143dda06f9/src/narration.ts#L646)):*

No. We're typically trying to scrollIntoView a paragraph whose parent is a bloom-editable which (being in a bubble) may be partly off-screen in any direction. It may also be possible that we're trying to scrollIntoView a span whose parent is a paragraph inside a BE. So left:0 is the far left of the Bloom-editable (or maybe the paragraph), which (with centered text especially) may be a bit further left than the first visible character.
</blockquote></details>

Ok, thanks. Sorry; I should have looked (and thought) more carefully.

I just kept thinking we should be able to come up with an algorithm which doesn't have a problem with this pathological case -- and I'm sure we could -- but you were right that it isn't worth bothering.

PR #288, opened 2022-10-24
Work around Firefox 60 bug/feature to prevent unwanted narration (BL-11449)

StephenMcConnel:
- - -
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/288)

### DIFF
--- a/src/narration.ts
+++ b/src/narration.ts
@@ -880,48 +880,76 @@ export default class Narration {
               );
     }
 
-    // Optional tryHarder param indicates that we may be called before the DOM is
-    // displayed, so we can't depend on the simple testing of display.
-    private getPlayableDivs(container: HTMLElement, tryHarder?: boolean) {
+    // When changing the language on a page, this function may be called before the DOM is
+    // fully displayed, so we can't depend on simply testing display to determine whether
+    // an editable div is visible and thus a playable div.  Firefox 60 initializes
+    // display="block" for all editable divs while newer browsers initialize it to null.
+    // The optional isForPlayButtonDisplay indicates (if true) that all the caller wants is
+    // a non-empty array if the Play button is supposed to be displayed and the divs returned
+    // will not actually be played.
+    private getPlayableDivs(
+        container: HTMLElement,
+        isForPlayButtonDisplay?: boolean
+    ) {
         // We want to play any audio we have from divs the user can see.
         // This is a crude test, but currently we always use display:none to hide unwanted languages.
-        return this.findAll(".bloom-editable", container).filter(e => {
-            // When changing languages to display, this code gets called before the
-            // DOM has computed the display style for these elements.  Note that
-            // eliminating the tryHarder branch causes the Play button to disappear
-            // until the user changes pages, and automatically doing the tryHarder
-            // branch causes the audio to automatically play in the new language,
-            // but without any highlighting.
+        const editables = this.findAll(".bloom-editable", container);
+        if (
+            navigator.userAgent.indexOf("Firefox/60.0") &&
+            // Firefox 60 initializes display="block" everywhere instead of display=null.
+            // So, we check whether display="block" is set in all editable divs and return
+            // an empty array if we're retrieving divs to actually play.
+            // The normal rules are followed to allow the play button to show.  Of course,
+            // if there isn't any audio on this page, it may show the play button when
+            // it shouldn't, but working with obsolete software, one can't have everything...
+            !isForPlayButtonDisplay &&
+            !editables.some(e => window.getComputedStyle(e).display !== "block")
+        ) {
+            // All of them are 'block', so nothing has been settled yet for actual visibility
+            // and thus playability.  Things will work okay if the user decides to click on
+            // the Play button later since the display="none" styles will have been set by then.
+            return [];
+        }
+        return editables.filter(e => {
             const display = window.getComputedStyle(e).display;
             if (display && display !== "none") return true;
+            // We want the div for the current language even if display has not yet been set,
+            // but only in the case where we're trying to determine whether to display the
+            // Play button.  (Note, Firefox 60 never reaches this line in the code.)
             return (
-                tryHarder && !display && e.lang && e.lang === this.currentLang
+                isForPlayButtonDisplay &&
+                !display &&
+                e.lang &&
+                e.lang === this.currentLang
             );
         });
     }
 
     // Optional page param is for use when 'playerPage' has NOT been initialized.
     // Not using the optional param assumes 'playerPage' has been initialized
-    // Optional tryHarder param indicates that we may be called before the DOM is
-    // displayed, so we can't depend on the simple testing of display.
+    // Optional isForPlayButtonDisplay param indicates that we want the results only
+    // to control the display of the Play button.
     private getPagePlayableDivs(
         page?: HTMLElement,
-        tryHarder?: boolean
+        isForPlayButtonDisplay?: boolean
     ): HTMLElement[] {
-        return this.getPlayableDivs(page ? page : this.playerPage, tryHarder);
+        return this.getPlayableDivs(
+            page ? page : this.playerPage,
+            isForPlayButtonDisplay
+        );
     }
 
     // Optional page param is for use when 'playerPage' has NOT been initialized.
     // Not using the optional param assumes 'playerPage' has been initialized
-    // Optional tryHarder param indicates that we may be called before the DOM is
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
Sorry, finding this pretty confusing. If my suggestions just confuse you, maybe we should meet?

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/288#-:-NFAq0-L7hD4Ni05oQvQ:b-dthdba)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/narration.ts` line 894 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/288#-NFAc4OcEvlpVPf3tyT5:-NFAc4OdEHHIFxKaUYMr:b-46nh7f) ([raw file](https://github.com/bloombooks/bloom-player/blob/69bb88d7bbe803d0676f7f6a60a512218c0908fb/src/narration.ts#L894)):*
> ```XML
>         const blocks = editables.filter(e => {
>             const display = window.getComputedStyle(e).display;
>             return display === "block";
> ```

We only use 'blocks' in a condition involving its length. But then that condition is independent of the display value we compute and of the list member. If that condition is true, every member of the list will be filtered out. So why not just have the whole special case here and test it once? Something like this,
if (navigator.userAgent.indexOf("Firefox/60.0") &&
  domIsStable && // !tryHarder, or should it be !domIsStable? see below
  editable.length > 1 && // why does it take TWO to make a special case? If there's only one assume it must be visible? Comment please.
  ! editable.some(e => window.getComputedStyle(e).display !== 'block')) { // i.e., all of them are 'block'
    return [];
}
And then you can leave the loop for the normal case unchanged.
___
*[`src/narration.ts` line 909 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/288#-NFAeP9SCLsAj1C-DUVb:-NFAeP9SCLsAj1C-DUVc:bqy44h5) ([raw file](https://github.com/bloombooks/bloom-player/blob/69bb88d7bbe803d0676f7f6a60a512218c0908fb/src/narration.ts#L909)):*
> ```XML
>                 // So, we check whether display=block is set in multiple divs and return
>                 // false for the filter when getting divs to actually play, but we follow
>                 // the normal rules to allow the play button to show otherwise.  Of course,
> ```

These two lines of comment suggest that a better name for 'tryharder' is possible. Seems like we need to 'tryharder' when the display hasn't had time to stabilize (=enabling play button?) but should not (for better reliability?) if we think it's stable. So would it be valid to rename 'tryharder' to 'domMayNotBeStable'? Or reverse its sense to avoid double negatives and call it 'domIsStable' or 'trustDisplay{For}Visibility'?
But then the condition below feels backwards. !tryharder = 'dom is stable' but then it's when we're told the DOM IS stable that we do the special case, returning false if every div is display:block (and there are at least two)
___
*[`src/narration.ts` line 923 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/288#-NFAlNPj6VrLjI4J-M0U:-NFAlNPj6VrLjI4J-M0V:bbpwgc) ([raw file](https://github.com/bloombooks/bloom-player/blob/69bb88d7bbe803d0676f7f6a60a512218c0908fb/src/narration.ts#L923)):*
> ```XML
>             if (display && display !== "none") return true;
>             return (
>                 tryHarder && !display && e.lang && e.lang === this.currentLang
> ```

!display means that it's null, right? Which means things are not stable, because if they were, we would have figured out some definite value for display. So maybe tryharder/!domIsStable is redundant here?
Would it be clearer (with the same result) like this?
if (!display) { // && tryHarder or && !domIsStable if necessary?
   // things are not stable (and we're not running Gecko60)...approximate visibility based on lang
   return e.lang && e.lang === this.currentLang
}
return display !== 'none'; // normal case, true if it's visible.

StephenMcConnel (commented):
I was confused while figuring out what was going on, so I was just sharing the confusion.  :-)  Stepping back and rethinking a bit, I think I've provided better commentary and a better variable name.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/288#-:-NFB9luXAmazuVhUhBWQ:b-3bxi0n)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/narration.ts` line 894 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/288#-NFAc4OcEvlpVPf3tyT5:-NFAzJj1AvdQBuuFdsGu:b-uwv7ub) ([raw file](https://github.com/bloombooks/bloom-player/blob/69bb88d7bbe803d0676f7f6a60a512218c0908fb/src/narration.ts#L894)):*

We only use 'blocks' in a condition involving its length. But then that condition is independent of the display value we compute and of the list member. If that condition is true, every member of the list will be filtered out. So why not just have the whole special case here and test it once? Something like this,
if (navigator.userAgent.indexOf("Firefox/60.0") &&
  domIsStable && // !tryHarder, or should it be !domIsStable? see below
  editable.length > 1 && // why does it take TWO to make a special case? If there's only one assume it must be visible? Comment please.
  ! editable.some(e => window.getComputedStyle(e).display !== 'block')) { // i.e., all of them are 'block'
    return [];
}
And then you can leave the loop for the normal case unchanged.
</blockquote></details>

I moved this test outside the filter loop as suggested.
Testing for length > 2 in addition to equal lengths is probably overkill.  I was accounting for the lang="z" divs, but don't really need to do so.
___
*[`src/narration.ts` line 909 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/288#-NFAeP9SCLsAj1C-DUVb:-NFAywjo4Y3Tff_qnpZ-:byv2r2k) ([raw file](https://github.com/bloombooks/bloom-player/blob/69bb88d7bbe803d0676f7f6a60a512218c0908fb/src/narration.ts#L909)):*

These two lines of comment suggest that a better name for 'tryharder' is possible. Seems like we need to 'tryharder' when the display hasn't had time to stabilize (=enabling play button?) but should not (for better reliability?) if we think it's stable. So would it be valid to rename 'tryharder' to 'domMayNotBeStable'? Or reverse its sense to avoid double negatives and call it 'domIsStable' or 'trustDisplay{For}Visibility'?
But then the condition below feels backwards. !tryharder = 'dom is stable' but then it's when we're told the DOM IS stable that we do the special case, returning false if every div is display:block (and there are at least two)
</blockquote></details>

tryHarder is true when we want the result for whether or not to display the Play button rather than for actually playing.  I hadn't yet come up with a good name (or apparently a good explanation) yet.  I renamed the variable to 'isForPlayButtonDisplay' and edited the comments extensively.
___
*[`src/narration.ts` line 923 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/288#-NFAlNPj6VrLjI4J-M0U:-NFB7rWDELIkf6XwOdxI:b-8li688) ([raw file](https://github.com/bloombooks/bloom-player/blob/69bb88d7bbe803d0676f7f6a60a512218c0908fb/src/narration.ts#L923)):*

!display means that it's null, right? Which means things are not stable, because if they were, we would have figured out some definite value for display. So maybe tryharder/!domIsStable is redundant here?
Would it be clearer (with the same result) like this?
if (!display) { // && tryHarder or && !domIsStable if necessary?
   // things are not stable (and we're not running Gecko60)...approximate visibility based on lang
   return e.lang && e.lang === this.currentLang
}
return display !== 'none'; // normal case, true if it's visible.
</blockquote></details>

The specification in the issue said not to play any sounds when changing the language.  So I think the logic here is correct, it just needs a better variable name to make sense.  (along with a better explanation, which I think I've provided)

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/288)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

___
*[`src/narration.ts` line 894 at r1](https://reviewable.io/reviews/bloombooks/bloom-player/288#-NFAc4OcEvlpVPf3tyT5:-NFEOvBZ7WmUMltgpDxY:bo8squn) ([raw file](https://github.com/bloombooks/bloom-player/blob/69bb88d7bbe803d0676f7f6a60a512218c0908fb/src/narration.ts#L894)):*

I moved this test outside the filter loop as suggested.
Testing for length > 2 in addition to equal lengths is probably overkill.  I was accounting for the lang="z" divs, but don't really need to do so.
</blockquote></details>

I think those ought to be filtered out...no reason to include them in a BloomPub

PR #291, opened 2022-11-11
Set up activities before their first display (BL-11449)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/291)

### DIFF
--- a/src/activities/activityManager.ts
+++ b/src/activities/activityManager.ts
@@ -14,8 +14,15 @@ export interface IActivityModule {
 
 // This is the class that the activity module has to implement
 export interface IActivityObject {
-    prepare: (context: ActivityContext) => void;
+    // Do one-time initialization of the page's html. In the current implementation,
+    // this is acting on a copy of the html, not the real DOM which the user will interact with.
+    // So we can't do things like set up event handlers. (See showingPage below.)
+    initializePageHtml: (context: ActivityContext) => void;
+
+    // Do any setup needed each time the activity becomes the active page.
+    // This is acting on the real DOM, so this is the time to set up event handlers, etc.
     showingPage: (context: ActivityContext) => void;
+
     stop: () => void;
 }
 // Constructing stuff from interfaces has problems with typescript at the moment.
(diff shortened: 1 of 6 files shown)


andrew-polk:
The first four commits are just backing out all the other changes we have made for this card so far.
I've already set Reviewable to be by commit, and I think I've made it so a reviewer will only see the final commit (with the prior ones already being approved). But you may still need to skip through the first four commits yourself.

andrew-polk:
Note that we are treating the fact that changing languages while on an activity page means it doesn't get the click handlers set up as a separate bug, not dealt with in this PR.

JohnThomson (changes requested):
This is difficult. Some stuff just got refactored into new methods (nice, but makes it hard to see what changed). Some stuff got moved to the initial creation of the page list. It looks plausible, but I can't really be sure whether anything got lost or added along the way. Is there anything I should particularly attend to?

I do think it would be helpful to capture in comments WHY some activity initialization is best done during initial setup of the whole book, and some must be done only when the page is shown. (In particular, that anything that attaches events to the DOM can't be done during page list creation, because it's not working on the real DOM.)

> Reviewed 1 of 2 files at r3, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/291#-:-NGd-vVH4-vtZPKHsUxL:bu0db39)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/activities/activityManager.ts` line 179 at r5](https://reviewable.io/reviews/bloombooks/bloom-player/291#-NGd1fsV6j_hOHDElnFl:-NGd1fsV6j_hOHDElnFm:b-kivydr) ([raw file](https://github.com/bloombooks/bloom-player/blob/343262afed59badd0764d078e104ca8a0182aae5/src/activities/activityManager.ts#L179)):*
> ```XML
>             ) as IActivityObject;
>             const activityContext = this.getActivityContext(pageIndex, pageDiv);
>             activityPage.prepare(activityContext);
> ```

If I'm reading it right, here we are running code that came from the activity, and when we constructed the object that has the prepare method, we passed it the page. If so, how do we know THAT code doesn't count on being able to add events to what it assumes is the live DOM element? Or capture a reference to it in some variable? Because I think where you used to do this, the activity code was getting passed the real DOM element, but now it's just the temporary one we parsed from the HTML file and will later convert back to a string to pass to dangerouslySetInnterHtml.

andrew-polk (commented):
I don't think there is anything specific to look at other than what you've already listed. 

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/291#-:-NI-TlA82rGElvMt5zHI:bmxvbm)* status: 1 of 7 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/activities/activityManager.ts` line 179 at r5](https://reviewable.io/reviews/bloombooks/bloom-player/291#-NGd1fsV6j_hOHDElnFl:-NI-M3GD2VU1OBNP9rWU:byh98l5) ([raw file](https://github.com/bloombooks/bloom-player/blob/343262afed59badd0764d078e104ca8a0182aae5/src/activities/activityManager.ts#L179)):*

If I'm reading it right, here we are running code that came from the activity, and when we constructed the object that has the prepare method, we passed it the page. If so, how do we know THAT code doesn't count on being able to add events to what it assumes is the live DOM element? Or capture a reference to it in some variable? Because I think where you used to do this, the activity code was getting passed the real DOM element, but now it's just the temporary one we parsed from the HTML file and will later convert back to a string to pass to dangerouslySetInnterHtml.
</blockquote></details>

I think you're right about that change (which I hadn't realized until you brought it up).
Thankfully, the only real implementer of `prepare` is the one we are trying to fix, and it is only making html changes which we **want** in the base html (and can therefore operate on a copy).

I've attempted to make this more obvious by renaming `prepare` to `prepareBaseHtml` and added some further clarifying comments. 
See what you think. Open to better naming, etc.

JohnThomson (approved):
LGTM, though I'd like to improve on the 'prepareBaseHtml' name if we can, so I won't merge until you have a chance to think about that.

> Reviewed 5 of 6 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/291#-:-NIrukRi4hsQuP2c2_Qg:b-elxg6)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/activities/activityManager.ts` line 20 at r7](https://reviewable.io/reviews/bloombooks/bloom-player/291#-NIrtQecDgEesw1Zuza7:-NIrtQecDgEesw1Zuza8:b9btywa) ([raw file](https://github.com/bloombooks/bloom-player/blob/fb320f52907798fbb53c99955c6ed8ed2abadd98/src/activities/activityManager.ts#L20)):*
> ```XML
>     // this is acting on a copy of the html, not the real DOM which the user will interact with.
>     // So we can't do things like set up event handlers. (See showingPage below.)
>     prepareBaseHtml: (context: ActivityContext) => void;
> ```

Seeing this in a diff, I had a hard time figuring out the difference between prepareBaseHtml and initializePageHtml. It would probably be less of a problem in the full context. Still, if I'm understanding right, they are run at the same time (actually the latter calls the former) with the same restrictions; the difference is that one is the activityObject method and the other belongs to activityManager. But they are similar enough that I would be inclined to give them the SAME name, unless you see some reason that would be confusing. I prefer initializePageHtml; it isn't obvious in what sense the HTML being prepared is a 'base' for something.

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloom-player/291)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/activities/activityManager.ts` line 20 at r7](https://reviewable.io/reviews/bloombooks/bloom-player/291#-NIrtQecDgEesw1Zuza7:-NIs90o1203zp0DpuW_u:b-c1gesv) ([raw file](https://github.com/bloombooks/bloom-player/blob/fb320f52907798fbb53c99955c6ed8ed2abadd98/src/activities/activityManager.ts#L20)):*

Seeing this in a diff, I had a hard time figuring out the difference between prepareBaseHtml and initializePageHtml. It would probably be less of a problem in the full context. Still, if I'm understanding right, they are run at the same time (actually the latter calls the former) with the same restrictions; the difference is that one is the activityObject method and the other belongs to activityManager. But they are similar enough that I would be inclined to give them the SAME name, unless you see some reason that would be confusing. I prefer initializePageHtml; it isn't obvious in what sense the HTML being prepared is a 'base' for something.
</blockquote></details>

Yeah, I struggled. But after weighing it all, I think you're right.

PR #298, opened 2023-01-13
Fix bloom-player pages/buttons misaligned (BL-11497)

sujeffreyl:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/298)

### DIFF
--- a/src/bloom-player-controls.tsx
+++ b/src/bloom-player-controls.tsx
@@ -19,7 +19,8 @@ import React, { useState, useEffect, useRef, LegacyRef } from "react";
 import LangData from "./langData";
 import {
     getQueryStringParamAndUnencode,
-    getBooleanUrlParam
+    getBooleanUrlParam,
+    getNumericUrlParam
 } from "./utilities/urlUtils";
 import { IconButton } from "@material-ui/core";
 //tslint:disable-next-line:no-submodule-imports
@@ -31,7 +32,7 @@ import { withStyles, createTheme } from "@material-ui/core/styles";
 import DragBar from "@material-ui/core/Slider";
 import { bloomRed } from "./bloomPlayerTheme";
 import { setDurationOfPagesWithoutNarration } from "./narration";
-import { roundToNearestEven } from "./utilities/mathUtils";
+import { roundToNearestK } from "./utilities/mathUtils";
 
 // This component is designed to wrap a BloomPlayer with some controls
 // for things like pausing audio and motion, hiding and showing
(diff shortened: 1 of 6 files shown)


JohnThomson (commented):
Nice job tracking this down and coming up with a fix!
I don't think we should use 'host', which is intended just to provide identifying information for analytics, to trigger different BP behavior. So I'd like to see a new param used to control this.
From an experiment I suggested and you did, it seems that needing an even number is related to the fact that BE scales the whole BP by a factor of 2 (see pixelDensityMultiplier in BPs DeviceFrame.less). If it's not a lot of work, a param that could just be tweaked if the scaling changes would be ideal. But if that's a lot of work it may be a case of YAGNI, in which case, I'd be OK with keeping the current algorithm but controlling it with a new param like forceEvenWidth.
In either case, I think there should be a parallel change in BE to comment pixelDensityMultiplier noting that BP (or the param passed to it) may need to change if pixelDensityMultiplier does, and referencing this issue.
If this turns out to be more than a couple of hours' work, maybe get another opinion or discuss it some more!

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/298#-:-NM-nrSE3bgDpZptOXtg:b-knjr7h)* status: 0 of 5 files reviewed, all discussions resolved

sujeffreyl (commented):
It is sadly not as simple as pixelDensityMultiplier :(

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/298#-:-NMZfSgV-QxD4-kZpHDJ:b-j8ovqr)* status: 0 of 6 files reviewed, all discussions resolved

JohnThomson (changes requested):
You've done a lot more work on this. I hope it was worth it. One quibble remaining...

> Reviewed 2 of 5 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/298#-:-NMa3w48-v9HV_6KNq7p:b-3267yx)* status: all files reviewed, 1 unresolved discussion (waiting on @sujeffreyl)

___
*[`src/utilities/urlUtils.ts` line 7 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/298#-NMa3-po6OuQvJELIcXB:-NMa3-po6OuQvJELIcXC:b-mgrzo2) ([raw file](https://github.com/BloomBooks/bloom-player/blob/ffc30fa4cf67ff88943cce47dbb8bade76481aa0/src/utilities/urlUtils.ts#L7)):*
> ```XML
> // but for now just leaving this function definition to wrap from a type-checked call
> // to a non-checked call.
> export function getQueryStringParamAndUnencode(
> ```

What's with the Unencode? It's not obvious that any unencoding is (or was) being done.

sujeffreyl (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/298)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/utilities/urlUtils.ts` line 7 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/298#-NMa3-po6OuQvJELIcXB:-NMaE0vL7daehVW0irP6:bvy3osn) ([raw file](https://github.com/BloomBooks/bloom-player/blob/ffc30fa4cf67ff88943cce47dbb8bade76481aa0/src/utilities/urlUtils.ts#L7)):*

What's with the Unencode? It's not obvious that any unencoding is (or was) being done.
</blockquote></details>

That's the old name for it, created by JH.   The decoding is done further down in the stack trace. parseUriComponent() will get called which will call decodeURIComponent() on stuff.

JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/298)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sujeffreyl)

PR #301, opened 2023-04-04
Stop old audio playback when language changed (BL-10511)

jsubloom:
Also updates the visibility of the PlayPause button

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)

### DIFF
--- a/src/bloom-player-controls.tsx
+++ b/src/bloom-player-controls.tsx
@@ -866,9 +866,7 @@ export const BloomPlayerControls: React.FunctionComponent<IProps &
                 showPlayPause={hasAudio || hasMusic || hasVideo}
                 bookLanguages={languageData}
                 activeLanguageCode={activeLanguageCode}
-                onLanguageChanged={(isoCode: string) =>
-                    handleLanguageChanged(isoCode)
-                }
+                onLanguageChanged={handleLanguageChanged}
                 canShowFullScreen={!props.hideFullScreenButton}
                 extraButtons={props.extraButtons}
                 bookHasImageDescriptions={bookHasImageDescriptions}
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -943,11 +943,18 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
             }, 500);
         } else {
             if (BloomPlayerCore.currentPage) {
-                // We need to replace the old currentPage with the corresponding one created from the updated content.
                 window.setTimeout(() => {
-                    BloomPlayerCore.currentPage = this.getPageAtSwiperIndex(
-                        BloomPlayerCore.currentPageIndex
-                    );
+                    // Presumably, the language was just changed.
+                    // Reset the index / re-show the page
+                    // Even though we're setting it to the same index, setIndex and showingPage have other useful/necessary side effects such as:
+                    // * Replace BloomPlayerCore.currentPage with the new page created from the updated content
+                    // * Update the Play/Pause button visibility based on the new language
+                    // * Stops the playback of the old language's audio (if in Play mode)
+                    // * Starts the playback of the new language's audio (if in Play mode)
+                    // * Update whether scrollbars should appear
+                    // * etc.
+                    this.setIndex(BloomPlayerCore.currentPageIndex);
+                    this.showingPage(BloomPlayerCore.currentPageIndex);
                 }, 200);
             }
         }
--- a/src/languageMenu.tsx
+++ b/src/languageMenu.tsx
@@ -15,7 +15,7 @@ import LangData from "./langData";
 
 interface ILanguageMenuProps {
     languages: LangData[];
-    onClose: (value: string) => void;
+    onClose: (value: string) => void; // Handler that runs when the menu closes. ${value} is the ISO code of the language the user selected, or "" if no change was made.
 }
 
 export const LanguageMenu: React.FunctionComponent<ILanguageMenuProps> = props => {


andrew-polk (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: all files reviewed, 1 unresolved discussion (waiting on @jsubloom)

___
*[`src/bloom-player-core.tsx` line 954 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/301#-NSD22YV2Fe_Vo9af4Kb:-NSD22YV2Fe_Vo9af4Kc:bti97ke) ([raw file](https://github.com/BloomBooks/bloom-player/blob/05e63826030edff2f910e124c4ccb7af9489a60e/src/bloom-player-core.tsx#L954)):*
> ```XML
>                     // When the language changes, we also want to update
>                     // whether the page has audio, music, video, etc.
>                     // so that BloomPlayerControls can update whether ${showPlayPause} should be true or not.
> ```

Does showingPage not get called in this case?
Seems like it should be, so if that can't be relied on, should we comment here about that?

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/bloom-player-core.tsx` line 954 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/301#-NSD22YV2Fe_Vo9af4Kb:-NSH3GKq74T1_-oEaXqP:b-27f9ma) ([raw file](https://github.com/BloomBooks/bloom-player/blob/05e63826030edff2f910e124c4ccb7af9489a60e/src/bloom-player-core.tsx#L954)):*

Does showingPage not get called in this case?
Seems like it should be, so if that can't be relied on, should we comment here about that?
</blockquote></details>

Well, it definitely doesn't get called in this case, or else we wouldn't have this issue reported. :) (Experimentally verified before as well)

So, an alternative solution is to call this.showingPage() right here instead.  I hesitated before because that would run a lot more code than the code I added, and I wasn't sure if that would have any unintended effects or not.
But I tried calling showingPage(...) and it seems to work as far as I can tell.

Shall we give that a try instead?

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: all files reviewed, 1 unresolved discussion (waiting on @jsubloom)

___
*[`src/bloom-player-core.tsx` line 954 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/301#-NSD22YV2Fe_Vo9af4Kb:-NSHF35w6Dh1CpzfEt5R:bbd6ni8) ([raw file](https://github.com/BloomBooks/bloom-player/blob/05e63826030edff2f910e124c4ccb7af9489a60e/src/bloom-player-core.tsx#L954)):*
<details><summary><i>Previously, jsubloom wrote…</i></summary><blockquote>

Well, it definitely doesn't get called in this case, or else we wouldn't have this issue reported. :) (Experimentally verified before as well)

So, an alternative solution is to call this.showingPage() right here instead.  I hesitated before because that would run a lot more code than the code I added, and I wasn't sure if that would have any unintended effects or not.
But I tried calling showingPage(...) and it seems to work as far as I can tell.

Shall we give that a try instead?
</blockquote></details>

I'm really torn. I wish I understood **why** we don't call it in this case. Like you, I would be very hesitant to add the extra code if I couldn't tell it was necessary.
Bloom player core is such a mess and feels so fragile.
Question, is... are we more likely to introduce bugs by calling it? Or are we more likely to fix future (and maybe even present) bugs by calling it?
I don't have a good answer. I'm ok with just adding a comment. I think I would be ok with calling `showingPage`, but then we have to answer, why just for this path? I.e. why not if it is a new book, etc.?

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/bloom-player-core.tsx` line 954 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/301#-NSD22YV2Fe_Vo9af4Kb:-NSM2Z5_77Bk23BdZRdW:bmoflxe) ([raw file](https://github.com/BloomBooks/bloom-player/blob/05e63826030edff2f910e124c4ccb7af9489a60e/src/bloom-player-core.tsx#L954)):*

I'm really torn. I wish I understood **why** we don't call it in this case. Like you, I would be very hesitant to add the extra code if I couldn't tell it was necessary.
Bloom player core is such a mess and feels so fragile.
Question, is... are we more likely to introduce bugs by calling it? Or are we more likely to fix future (and maybe even present) bugs by calling it?
I don't have a good answer. I'm ok with just adding a comment. I think I would be ok with calling `showingPage`, but then we have to answer, why just for this path? I.e. why not if it is a new book, etc.?
</blockquote></details>

For part of my investigation, I wonder: When we added handleLanguageCode, why did we tap into finishUp() with a new mode for finishUp()?  (And also, what do we use finishUp(false) for now?)

Initially:    Various code related to setting up the "pages" state (AKA setting the swiperContnet)

Now:    Still all the same "pages" state (AKA swiperContent) stuff.
Also:     showOrHideTitle2AndL1OnlyText(...)
             initializePageHtml(...)             
             this.props.pageStylesAreNowInstalled();
             BloomPlayerCore.currentPage = this.getPageAtSwiperIndex(BloomPlayerCore.currentPageIndex);

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/bloom-player-core.tsx` line 954 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/301#-NSD22YV2Fe_Vo9af4Kb:-NSMFKRmCnV6HgCQOr0O:bn2a7da) ([raw file](https://github.com/BloomBooks/bloom-player/blob/05e63826030edff2f910e124c4ccb7af9489a60e/src/bloom-player-core.tsx#L954)):*
<details><summary><i>Previously, jsubloom wrote…</i></summary><blockquote>

For part of my investigation, I wonder: When we added handleLanguageCode, why did we tap into finishUp() with a new mode for finishUp()?  (And also, what do we use finishUp(false) for now?)

Initially:    Various code related to setting up the "pages" state (AKA setting the swiperContnet)

Now:    Still all the same "pages" state (AKA swiperContent) stuff.
Also:     showOrHideTitle2AndL1OnlyText(...)
             initializePageHtml(...)             
             this.props.pageStylesAreNowInstalled();
             BloomPlayerCore.currentPage = this.getPageAtSwiperIndex(BloomPlayerCore.currentPageIndex);
</blockquote></details>

Q: Will we fix any future/present bugs by calling it?
A: Yes. Currently, if the page in the new lang needs scrollbars, it is errorneously not added.

Q:  I think I would be ok with calling showingPage, but then we have to answer, why just for this path? I.e. why not if it is a new book, etc.?
A1: Tautological/unsatistfying answer - because showingPage() is obviously/clearly called via some mechanism or another whenever you get a newBook.
A2: This might be more compelling for why we should add it in this path.  Actually, the newBookPath does end up calling showingPage() as well. See line 928 here: https://github.com/BloomBooks/bloom-player/blame/master/src/bloom-player-core.tsx

jsubloom (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @andrew-polk)

___
*[`src/bloom-player-controls.tsx` line 704 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/301#-NSMLXBw2J7fB97yzrcp:-NSMLXBw2J7fB97yzrcq:b-dl29t3) ([raw file](https://github.com/BloomBooks/bloom-player/blob/05e63826030edff2f910e124c4ccb7af9489a60e/src/bloom-player-controls.tsx#L704)):*
> ```XML
>         // But, if it ever were, I'm not so sure that ${paused} should always be set to true.
>         // Sometimes the book is supposed to start playing immediately upon load, no?
>         setPaused(true);
> ```

I changed my interpretation of what JH is asking for to be that if the old audio was currently playing, it should stop the old lang's audio and also start the audio in the new lang.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: all files reviewed, 1 unresolved discussion

andrew-polk (approved):
> Dismissed @jsubloom from a discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/301)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jsubloom)

PR #302, opened 2023-05-02
Handle RTL books (by reversing pages) (BL-8063)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/302)

### DIFF
--- a/src/bloom-player-controls.tsx
+++ b/src/bloom-player-controls.tsx
@@ -2,7 +2,7 @@
 BloomPlayerControls wraps BloomPlayerCore and adds just enough controls to preview the
 book inside of the Bloom:Publish:Android screen.
 */
-import { BloomPlayerCore } from "./bloom-player-core";
+import { BloomPlayerCore, adjustStartPage } from "./bloom-player-core";
 import * as ReactDOM from "react-dom";
 import {
     onBackClicked,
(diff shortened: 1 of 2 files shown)


andrew-polk (commented):
Did you look into https://swiperjs.com/swiper-api#method-swiper-changeLanguageDirection ?
Demo at https://swiperjs.com/demos#rtl

(might require an upgrade?)

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/302#-:-NUZ3YQEBJcm3N3IKDy5:b9ckf3j)* status: 0 of 2 files reviewed, all discussions resolved

JohnThomson (commented):
I did not discover this.
A quick experiment indicates that changeLanguageDirection is not a function on our version of Swiper. However, the demo you provided does not use it. As far as I can determine, the RTL behavior is produced simply by having a dir="rtl" attribute on the root swiper div. So I tried doing that manually with the inspector.
The pages do indeed slide the way we want.
However, the next/prev buttons are not reversed, and the direction of the page slider is not changed (i.e., the red blob for page 1 is towards the left, and page 2 is right of it). And it's not obvious how to get the dir attribute onto an element that Swiper controls. So there would be non-trivial work involved in redoing it this way.
Do you think it is worth it? Some things in the code would be simpler, not sure whether others might not be more complicated.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/302#-:-NUaq4hcBvpCnHoWrfmL:btubcc)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @andrew-polk)

JohnThomson:
Keeping this as a draft until we verify that what I think is a better approach works.

andrew-polk:
Superceded by #303

PR #307, opened 2024-02-23
BL-13053 fix talking book highlighting color customization

nabalone:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/307)

### DIFF
--- a/src/bloom-player.less
+++ b/src/bloom-player.less
@@ -62,11 +62,20 @@ above the page so that page numbers will increment.*/
     display: inline-block;
     white-space: pre-wrap;
 }
-// Note: :not(classA, classB) syntax not supported by some browsers, notably FF60.
-// Use :not(classB):not(classB) syntax instead!!!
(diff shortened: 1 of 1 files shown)


nabalone (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/307)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/bloom-player.less` line 60 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/307#-NrM_k4A8LNp6_ItCt-W:-NrM_k4A8LNp6_ItCt-X:btqvej7) ([raw file](https://github.com/BloomBooks/bloom-player/blob/a26f3db1baefb8fae05ff5f411b06a9b75c11380/src/bloom-player.less#L60)):*
> ```Less
> // deal with the most common case of overlap (the book title on the cover) by just
> // treating the text as a rectangle. And it's fine that this will show this way in Chrome too:
> .Title-On-Cover-style
> ```

This is still necessary to prevent BL-7442

PR #333, opened 2024-10-24
Prevent swiping pages when clicking on video buttons (BL-14021)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/333)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -221,6 +221,24 @@ interface IPlayerPageOptions {
     hideNavigation?: boolean;
 }
 
+// capturing mouse event handler added to body to prevent clicks on certain
+// elements from being seen by Swiper. This is an independent function so that
+// it doesn't end up being added more than once, even if the code that adds it
+// runs more often.
+function handleInputMouseEvent(event: Event) {
+    if ((event.target as HTMLElement).closest(".videoControlContainer")) {
+        // Stop Swier from seeing events on these elements.
+        // Note: Swiper version 11 has a class "swiper-no-swiping" that I think can be used to
+        // achieve this, or configured with noSwipingClass or noSwipingSelector.
+        // But we're at too low a version to use that, and bringing in the latest
+        // version is non-trivial.
+        event.stopPropagation();
+        // We don't need to preventDefault to keep Swiper from moving the page,
+        // and we don't want to because I think it would stop the mousedown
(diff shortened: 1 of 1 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/333)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/bloom-player-core.tsx` line 229 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/333#-OA-Iv-c8R5YSzYejMTu:-OA-Iv-c8R5YSzYejMTv:bo3p0hm) ([raw file](https://github.com/BloomBooks/bloom-player/blob/c1b7463dc62f5ba65e6fb0da2a1c94166b833249/src/bloom-player-core.tsx#L229)):*
> ```XML
> // runs more often.
> function handleInputMouseEvent(event: Event) {
>     if ((event.target as HTMLElement).closest(".videoControlContainer")) {
> ```

Will this prevent swiping on the whole video? I was expecting that we would just be limiting to the buttons on the video. Might seem like a small difference if the video is small, but if it were full page?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/333)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/bloom-player-core.tsx` line 229 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/333#-OA-Iv-c8R5YSzYejMTu:-OA2oP8j9RxXPzakYhN_:bafpm0r) ([raw file](https://github.com/BloomBooks/bloom-player/blob/c1b7463dc62f5ba65e6fb0da2a1c94166b833249/src/bloom-player-core.tsx#L229)):*

Will this prevent swiping on the whole video? I was expecting that we would just be limiting to the buttons on the video. Might seem like a small difference if the video is small, but if it were full page?
</blockquote></details>

No. VideoControlContainer is a class applied to the div that surrounds each of the button SVGs. Swiper behavior is only interfered with on the buttons themselves.

hatton (approved):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/333)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

PR #339, opened 2024-11-14
fix: Prevent scrolling from also swiping (BL-14079)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/339)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -23,12 +23,12 @@ import {
     reportPlaybackComplete,
     sendMessageToHost,
     setAmbientAnalyticsProperties,
-    updateBookProgressReport,
+    updateBookProgressReport
 } from "./externalContext";
 import {
     clearSoundLog,
     reportSoundsLogged,
-    sendStringToBloomApi,
+    sendStringToBloomApi
 } from "./videoRecordingSupport";
 import LangData from "./langData";
 import Replay from "@material-ui/icons/Replay";
@@ -49,7 +49,7 @@ import "jquery.nicescroll";
 import { getQueryStringParamAndUnencode } from "./utilities/urlUtils";
 import {
     kLocalStorageDurationKey,
-    kLocalStorageBookUrlKey,
+    kLocalStorageBookUrlKey
 } from "./bloomPlayerAnalytics";
 import { autoPlayType } from "./bloom-player-controls";
 import {
@@ -72,7 +72,7 @@ import {
     hidingPage as hidingNarrationPage,
     pageHasAudio,
     setIncludeImageDescriptions,
-    playAllSentences,
+    playAllSentences
 } from "./narration";
 import { logSound } from "./videoRecordingSupport";
 import { playSoundOf } from "./dragActivityRuntime";
@@ -123,7 +123,7 @@ interface IProps {
         // Provides the client a way to switch page numbers. It would feel more natural and React-y to
         // just pass current page number to bloom-player-controls as a property, but I don't think regenerating the whole
         // bloom-player-core to switch page numbers is a great idea.
-        setCurrentPage?: (pageNo: number) => void,
+        setCurrentPage?: (pageNo: number) => void
     ) => void;
 
     // A callback the client may supply to be notified whether navigation buttons are being hidden for this page.
@@ -277,7 +277,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
         // notifications from narration.ts etc about duration etc.
         BloomPlayerCore.currentPagePlayer = this;
         this.legacyQuestionHandler = new LegacyQuestionHandler(
-            props.locationOfDistFolder,
+            props.locationOfDistFolder
         );
     }
 
@@ -292,7 +292,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
         loadErrorHtml: "",
         ignorePhonyClick: false,
         isFinishUpForNewBookComplete: false,
-        inPauseForced: false,
+        inPauseForced: false
     };
 
     // The book url we were passed as a URL param.
@@ -330,15 +330,15 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
         // To get this to fire consistently no matter where the focus is,
         // we have to attach to the document itself. No level of react component
         // seems to work (using the OnKeyDown prop). So we use good ol'-fashioned js.
-        document.addEventListener("keydown", (e) =>
-            this.handleDocumentLevelKeyDown(e),
+        document.addEventListener("keydown", e =>
+            this.handleDocumentLevelKeyDown(e)
         );
 
         // Clicking on any element that has a data-href attribute will be treated as a link.
         // This is the simplest way to handle such links that may be scattered on different
         // types of elements throughout the book.  See BL-13879.
-        document.addEventListener("click", (e) =>
-            this.handleDocumentLevelClick(e),
+        document.addEventListener("click", e =>
+            this.handleDocumentLevelClick(e)
         );
 
         // We only need to add these body-level listeners once.
@@ -390,7 +390,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
             return;
         }
         const targetElement = (e.target as HTMLElement).closest(
-            "[data-href]",
+            "[data-href]"
         ) as HTMLElement;
         if (targetElement && targetElement.attributes["data-href"]) {
             const href = targetElement.attributes["data-href"].value as string;
@@ -468,10 +468,10 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
     private repairFF60Offset() {
         try {
             let wrapper = document.getElementsByClassName(
-                "swiper-wrapper",
+                "swiper-wrapper"
             )[0] as HTMLElement;
             let current = document.getElementsByClassName(
-                "swiper-slide-active",
+                "swiper-slide-active"
             )[0] as HTMLElement;
             if (wrapper && current && this.msToContinueFF60RepairChecks > 0) {
                 let error =
@@ -486,7 +486,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                         paddingPx.length === 0
                             ? 0
                             : parseFloat(
-                                  paddingPx.substring(0, paddingPx.length - 2),
+                                  paddingPx.substring(0, paddingPx.length - 2)
                               );
                     wrapper.style.paddingLeft = padding + error / scale + "px";
                 }
@@ -506,7 +506,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
     // don't consume extra power forever doing this check.
     private msToContinueFF60RepairChecks = 0;
 
-    private handleDocumentLevelKeyDown = (e) => {
+    private handleDocumentLevelKeyDown = e => {
         if (e.key === "Home") {
             this.goToFirstPage();
             e.preventDefault();
@@ -561,12 +561,12 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                 if (slashIndex > encodedSlashIndex) {
                     filename = this.sourceUrl.substring(
                         slashIndex + 1,
-                        this.sourceUrl.length,
+                        this.sourceUrl.length
                     );
                 } else {
                     filename = this.sourceUrl.substring(
                         encodedSlashIndex + 3,
-                        this.sourceUrl.length,
+                        this.sourceUrl.length
                     );
                 }
                 // Note, The current best practice is actually to have the htm file always be "index.htm".
@@ -580,7 +580,7 @@ export class BloomPlayerCore extends React.Component<IProps, IState> {
                 this.urlPrefix = haveFullPath
                     ? this.sourceUrl.substring(
                           0,
-                          Math.max(slashIndex, encodedSlashIndex),
+                          Math.max(slashIndex, encodedSlashIndex)
                       )
                     : this.sourceUrl;
                 this.music.urlPrefix = this.urlPrefix;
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
I'm seeing a lot of issues here, but maybe there's just one fundamental thing I'm misunderstanding. Did you test with all the different ways you can scroll a NiceScroll area on a device, or at least an emulation of a touch device?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-:-OC-xagsE7aHTiRbfDrk:b-jl44r9)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/bloom-player-core.tsx` line 2822 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC-ujaw6fv4WS379aFd:-OC-ujaw6fv4WS379aFe:b-gxsx76) ([raw file](https://github.com/BloomBooks/bloom-player/blob/04bdab71d37baaad2fa5d39c4ad93c2006c188a7/src/bloom-player-core.tsx#L2822)):*
> ```XML
>         bloomPage.removeEventListener("pointerdown", this.listenForPointerDown); // only want one!
>         bloomPage.addEventListener("pointerdown", this.listenForPointerDown);
>         for (const eventName of ["mousemove", "pointermove", "touchmove"]) {
> ```

What will swiper do if all it sees is a pointerDown and pointerUp with a large X separation but no move events? Or does setPointerCapture prevent that adequately? I don't understand why we need to handle move but not up events if setPointerCapture doesn't solve the whole problem.
___
*[`src/bloom-player-core.tsx` line 2831 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC-hG3LBHOm7Taj6XXm:-OC-hG3LBHOm7Taj6XXn:b-ek0xoz) ([raw file](https://github.com/BloomBooks/bloom-player/blob/04bdab71d37baaad2fa5d39c4ad93c2006c188a7/src/bloom-player-core.tsx#L2831)):*
> ```XML
>                     capture: true
>                 }
>             );
> ```

When you're dealing with the exact same function like this, it's not necessary to remove first. AddEventListener will not add the same function with the same options twice (unlike .NET). (But it has to be the same function instance, not just the same code declared in a local loop variable.)
___
*[`src/bloom-player-core.tsx` line 2852 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC-h_Kz2wzsVrq_QtMx:-OC-h_Kz2wzsVrq_QtMy:b-jzq3sg) ([raw file](https://github.com/BloomBooks/bloom-player/blob/04bdab71d37baaad2fa5d39c4ad93c2006c188a7/src/bloom-player-core.tsx#L2852)):*
> ```XML
>             (e.target as HTMLDivElement)?.classList?.contains(
>                 "nicescroll-cursors"
>             )
> ```

The target is the most local thing clicked, typically a TextNode. I think you need something like e.target?.closest(".nicescroll-cursors"). (Unless nicescroll attaches this class to everything in it's area? In which case many of them won't be divs, so it's confusing to indicate otherwise, though classList?.contains will save you from crashing. But if target is a textNode it certainly won't have any classList itself; that's a property of HTMLElements, so this will never succeed for those. 
Or are we only concerned about an event on the scrollbar itself? I would have thought at least on a touch device you could scroll by dragging on the text content.
Or maybe Nicescroll makes an overlay? The old comment for listenForPointerDown only talks about the thumb, but you seem to be using the same algorithm to deal with a move anywhere in the area controlled by NiceScroll.
Do we need to be concerned about a swipe which begins in the nicescrolll area but extends beyond it? I'm becoming convinced that nearly always, the right way to handle dragging is to just handle 'down' events with long-term handlers. The 'down' handler decides whether we want to do our thing to this event sequence. If so, the down handler attaches appropriate handlers to move and up (always on document with capture), and they can do their thing unconditionally. Up also removes them.
As a fringe benefit, you can capture the mousedown x in a local variable and use it in the move/up handlers.
In this case move/up would still have to consider whether there's an x component, but would not have to consider the target.
Andrew mentioned that we already have two special cases for where we don't want dragging to switch pages. Can we combine the tests for whether mouse down is in a no-page-turn area and handle them all the same?
___
*[`src/bloom-player-core.tsx` line 2860 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC-k-4N0Rt_Mp6MBZls:-OC-k-4N0Rt_Mp6MBZlt:bk8bp8u) ([raw file](https://github.com/BloomBooks/bloom-player/blob/04bdab71d37baaad2fa5d39c4ad93c2006c188a7/src/bloom-player-core.tsx#L2860)):*
> ```XML
>                 e.stopPropagation();
>             }
>             // As a backup, which doesn't work all the time for unknown reasons, stop the
> ```

One hypothesis:
event 1: mousedown: no movement at all: propagates
event 2: move up and left: x movement: stopped
event 3: move straight up: propagates.
Swiper calculates total movement since mousedown: it has an x component from event 2: swipe starts.

If you keep this strategy, you may need to record the mousedown x and consider whether the total movement has an x component, or even whether any event since then has had one.
___
*[`src/bloom-player-core.tsx` line 2864 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC-k1To6vYREZleszHQ:-OC-k1To6vYREZleszHR:bezso7b) ([raw file](https://github.com/BloomBooks/bloom-player/blob/04bdab71d37baaad2fa5d39c4ad93c2006c188a7/src/bloom-player-core.tsx#L2864)):*
> ```XML
>             // horizontal component of the movement is not zero.
>             else if ((e as any).movementX !== 0) {
>                 e.stopPropagation();
> ```

This will stop mousemove events that NiceScroll wants if the NiceScroll handler is attached to a parent of e.target

andrew-polk (commented):
As a higher level issue, I think this is now the third type of element and the third set of code related to not swiping when mousing down.
I guess it must be more complicated than generalizing to something like, if the target is one of these, don't swipe?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-:-OC4_4t92MwG6-sPLrxO:byqczxq)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @JohnThomson and @StephenMcConnel)

JohnThomson (changes requested):
I feel like I'm being difficult and don't want to discourage you. You've made considerable progress with a very difficult task! But I'm still struggling to understand some aspects of how your change works, and we want it to be maintainable by any of us. Nicescroll seems to be using some words in ways I don't recognize. I think of cursor as another name for either the mouse pointer or the text insertion point. The latter doesn't seem relevant and the mouse pointer is always inside the div it is clicking. Neither moves along a rail, so it sounds like you're talking about part of the scroll bar...but the only moving part there is the thing I would call the thumb.
Also, I get (now) that the bug report is for a computer with mouse and screen, and am glad you've also tested on a tablet and phone emulators. But there are other combinations to consider, especially when you start using things like maxTouchPoints. That will be non-zero for a laptop with a touchscreen and also for at least some kinds of trackpad and possibly if a digitizer is connected, even if the user is actually using the mouse. Our solution should work for all such situations if possible. Mostly now I'm quibbling about comments, so hopefully we're getting close. But maxTouchPoints has me worried.

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-:-OC5m4VeCXxejPhIrsoP:b-wwdnnc)* status: all files reviewed, 3 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/bloom-player-core.tsx` line 2852 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC-h_Kz2wzsVrq_QtMx:-OC5hdOA-oB_G3pf4cj7:bg2iykz) ([raw file](https://github.com/BloomBooks/bloom-player/blob/04bdab71d37baaad2fa5d39c4ad93c2006c188a7/src/bloom-player-core.tsx#L2852)):*

The target we want is the div containing the cursor itself, but it doesn't hurt to use closest().  The touch device does scroll by touching in the text area.  Touching in the cursor area has no effect.  The touch events get handled by either niceScroll or Swiper, but not by both.  (tested in emulator and real device with BRalpha and blorg in Chrome)
</blockquote></details>

I'm surprised it worked reliably without "closest" because I would not expect the target to be Div. It's supposed to be the innermost thing clicked, so in a bloom editable it would usually be a 'p' and might be an <em> or <strong> or <span> or whatever, especially if audio recording is involved. I'd still be happier if we weren't casting to HTMLDivElement since it probably isn't, unless your testing indicates that I'm wrong and it somehow IS usually a Div.
Unless we really are only targeting the thumb part of the scroll bar and that is a div. But you said somewhere that's just an indicator.
___
*[`src/bloom-player-core.tsx` line 2836 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC5gn3Y7ckEOx7tiKgR:-OC5gn3Y7ckEOx7tiKgS:b-6uxhgy) ([raw file](https://github.com/BloomBooks/bloom-player/blob/662502238617f06ce776e9235d23a89bd601a5c5/src/bloom-player-core.tsx#L2836)):*
> ```XML
> 
>     // This method is attached to the pointermove event.  If the event was caused by a
>     // mouse and is in the NiceScroll thumb area, we need to stop it from propagating any
> ```

Do we only want to stop it in the thumb area, or also in the rest of the area controlled by NiceScroll?
___
*[`src/bloom-player-core.tsx` line 2876 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC5c2ac70CVcVHiKEZm:-OC5c2ad506KajaDfjg2:b-tn42cz) ([raw file](https://github.com/BloomBooks/bloom-player/blob/662502238617f06ce776e9235d23a89bd601a5c5/src/bloom-player-core.tsx#L2876)):*
> ```XML
>                 // Pointer capture affects mouse events as well as pointer events.
>                 ev.stopPropagation();
>             }
> ```

I'm not sure what you're getting at with the maxTouchPoints test. Why would you want to stop mouse events on a laptop only if it doesn't have a touch screen or touchpad? At least a quick search indicates that a laptop touch screen should have a non-zero maxTouchPoints. And likewise a touchpad (some of which support multiple touches). Then there are digitizer tablets...I think at least we need a comment explaining what the maxTouchPoints test is for. (Also above)

StephenMcConnel (commented):
There's nothing we can really guarantee with other combinations than a touchscreen device and a mouse device.  We just don't have the hardware to test on.  I decided to remove the tests for maxTouchPoints because they may not add anything except confusion.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-:-OCA3Jh-18v3Y0_BW1ZV:b-85hc6j)* status: all files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`src/bloom-player-core.tsx` line 2852 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC-h_Kz2wzsVrq_QtMx:-OC9ydh4-xsm7eNbmT3W:b5o2uqk) ([raw file](https://github.com/BloomBooks/bloom-player/blob/04bdab71d37baaad2fa5d39c4ad93c2006c188a7/src/bloom-player-core.tsx#L2852)):*

I'm surprised it worked reliably without "closest" because I would not expect the target to be Div. It's supposed to be the innermost thing clicked, so in a bloom editable it would usually be a 'p' and might be an <em> or <strong> or <span> or whatever, especially if audio recording is involved. I'd still be happier if we weren't casting to HTMLDivElement since it probably isn't, unless your testing indicates that I'm wrong and it somehow IS usually a Div.
Unless we really are only targeting the thumb part of the scroll bar and that is a div. But you said somewhere that's just an indicator.
</blockquote></details>

niceScroll inserts div elements outside the immediate hierarchy containing the actual text.  (They're inserted in parallel with the parent of the div.bloom-translationGroup.)  These are positioned absolutely to the side (or bottom) and the parent div.nicescroll-rails are given a z-index value so that they intercept the mouse click before the underlying text elements.  The child div.nicescroll-cursor is the active div which controls the display of the thumb.  It is positioned inside the div.nicescroll-rails with a fixed height and a top value corresponding to the scroll position (and thus the display of the thumb).  For mouse devices, scrolling doesn't start unless the mouse down is inside the area currently occupied by the div.nicescroll-cursor, at least not for displaying the thumb moving in sync with the mouse.
I agree that cursor was an unfortunate choice for using in the classname, and which I copied in earlier comments.  I've tried to change the comments in the code to use thumb throughout.
___
*[`src/bloom-player-core.tsx` line 2836 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC5gn3Y7ckEOx7tiKgR:-OC9yPaw1niJJWdncs70:b-q552o9) ([raw file](https://github.com/BloomBooks/bloom-player/blob/662502238617f06ce776e9235d23a89bd601a5c5/src/bloom-player-core.tsx#L2836)):*

Do we only want to stop it in the thumb area, or also in the rest of the area controlled by NiceScroll?
</blockquote></details>

The thumb area is sufficient.  Clicking the mouse down outside the thumb area doesn't start a scroll, at least not one that has the thumb display track the mouse movement.
___
*[`src/bloom-player-core.tsx` line 2876 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/339#-OC5c2ac70CVcVHiKEZm:-OCA0zAyCn9Ks59LEHPC:b-whppyv) ([raw file](https://github.com/BloomBooks/bloom-player/blob/662502238617f06ce776e9235d23a89bd601a5c5/src/bloom-player-core.tsx#L2876)):*

I'm not sure what you're getting at with the maxTouchPoints test. Why would you want to stop mouse events on a laptop only if it doesn't have a touch screen or touchpad? At least a quick search indicates that a laptop touch screen should have a non-zero maxTouchPoints. And likewise a touchpad (some of which support multiple touches). Then there are digitizer tablets...I think at least we need a comment explaining what the maxTouchPoints test is for. (Also above)
</blockquote></details>

I've removed the maxTouchPoints test in both places.  I decided that testing for pointerType === mouse is the best we can do.  If other hardware still has problems with this bug, we can't fix it.

JohnThomson (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/339)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #340, opened 2024-11-20
fix: Prevent showing game answers when sliding pages (BL-14064)

StephenMcConnel:
The only non-formatting change is the call to doShowAnswersInTargets which changes "true" to "!inPlayer" (plus the preceding line, of course) around line 297.  I don't why VSCode changed so much formatting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/340)


andrew-polk:
I ran the new prettier on all .ts and .tsx files then rebased this onto the change, so now it only shows the real change in this PR.

PR #349, opened 2024-12-23
Prevent various unwanted things in Bloom Player

hatton:
When clicking a link, do not toggle full-window mode
When doing something with a video, don't drag
When clicking on a fixed image, don't let the browser do that weird ghost drag thing with hit

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/349)


JohnThomson:
Added a commit (by request) fixing some problems I saw in code review.
JohnH should probably check those changes.

PR #351, opened 2025-01-14
fix: adjust page scaling when following link to book (BL-14198)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/351)

### DIFF
--- a/src/bloom-player-controls.tsx
+++ b/src/bloom-player-controls.tsx
@@ -976,6 +976,13 @@ export const BloomPlayerControls: React.FunctionComponent<BloomPlayerProps> = (
                     );
                     setPageNumbers(bookProps.pageNumbers);
                     setIsRtl(bookProps.isRtl);
+                    if (bookProps.internalUrl != currentCoreUrl) {
+                        if (currentCoreUrl) {
+                            // We're changing books, so we need to re-scale.
+                            rerunScalePageToWindow();
+                        }
+                        currentCoreUrl = bookProps.internalUrl;
+                    }
                 }}
                 controlsCallback={updateControlsWhenOpeningNewBook}
                 setForcedPausedCallback={(p) => {
@@ -1152,3 +1159,4 @@ export function InitBloomPlayerControls() {
         document.getElementById("root"),
     );
 }
+let currentCoreUrl = "";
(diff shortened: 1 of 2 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/351)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`src/bloom-player-controls.tsx` line 986 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/351#-OGao1mq-1mskC9pIwzs:-OGao1mq-1mskC9pIwzt:bqop1fn) ([raw file](https://github.com/BloomBooks/bloom-player/blob/b9f1c0934333debd02f362e1cfd43da8508d3abf/src/bloom-player-controls.tsx#L986)):*
> ```XML
>                     // follows a link to a different book, at which point we DO need to re-scale,
>                     // since that book may use a different page size or orientation.
>                     // So, do the rescale if there's been a previous notification and
> ```

If the URL changes within a second (e.g. e2e test), then we don't adjust? Should we say time OR URL change?

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/351)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`src/bloom-player-controls.tsx` line 986 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/351#-OGao1mq-1mskC9pIwzs:-OGbDrXPEE4Or82jq529:bep4shp) ([raw file](https://github.com/BloomBooks/bloom-player/blob/b9f1c0934333debd02f362e1cfd43da8508d3abf/src/bloom-player-controls.tsx#L986)):*

If the URL changes within a second (e.g. e2e test), then we don't adjust? Should we say time OR URL change?
</blockquote></details>

Don't need the time if we can tell that the URL changed. But that was internal to bloom-player-core. Reworked.

hatton (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/351)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

PR #357, opened 2025-02-05
feat: Export shared source files in bloom-player package (BL-14008)

StephenMcConnel:
The major work was merging the changes that have been made to these four files differently in bloom-player and BloomDesktop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/357)

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,11 +72,16 @@
         "vite-plugin-static-copy": "^2.0.0",
         "vitest": "^2.1.3"
     },
+    "// The four src/ files are directly exported so BloomDesktop can consume them as shared code.": "",
     "files": [
         "dist/*.js",
         "dist/bloomplayer.htm",
         "dist/*.mp3",
         "lib/*.js",
+        "src/event.ts",
+        "src/narration.ts",
+        "src/dragActivityRuntime.ts",
+        "src/scrolling.ts",
         "dist/*.css",
         "dist/*.tsv"
     ],
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -2264,7 +2264,7 @@ export class BloomPlayerCore extends React.Component<IProps, IPlayerState> {
             );
             return; // nothing more we can do until the page we want really exists.
         }
-        addScrollbarsToPage(bloomPage);
+        addScrollbarsToPage(bloomPage, BloomPlayerCore.handlePointerMoveEvent);
     }
 
     // This method is attached to the pointermove and pointerup events.  If the event was
--- a/src/dragActivityRuntime.ts
+++ b/src/dragActivityRuntime.ts
@@ -1,5 +1,6 @@
 // This is the code that is shared between the Play tab of the bloom games
-// (also known as drag activities) and bloom-player.
+// (also known as drag activities) and bloom-player.  The code originates in
+// the bloom-player repo and is shared by the bloom-player package.
 // It wants to live in the dragActivity folder because it is specific to drag activities.
 // However, it also wants to live in the same place relative to narration.ts both in
 // bloom player and bloom desktop. For now that's a stronger requirement.
@@ -55,6 +56,33 @@ const restorePositions = () => {
     positionsToRestore = [];
 };
 
+// This method may be specific to BloomDesktop.
+export function adjustDraggablesForLanguage(page: HTMLElement) {
+    if (page.getAttribute("data-activity") !== "drag-letter-to-target") {
+        return;
+    }
+    const draggables = Array.from(page.querySelectorAll("[data-bubble-id"));
+    draggables.shift(); // The first one is always visible.
+    draggables.forEach((draggable: HTMLElement) => {
+        const shouldBeVisible = !!(
+            draggable.getElementsByClassName(
+                "bloom-visibility-code-on",
+            )[0] as HTMLElement
+        )?.innerText.trim();
+        draggable.classList.toggle("bloom-unused-in-lang", !shouldBeVisible);
+        const target = getTarget(draggable);
+
+        if (target) {
+            target.classList.toggle("bloom-unused-in-lang", !shouldBeVisible);
+        }
+    });
+}
+export function IsRunningOnBloomDesktop(bloomPage?: Element): boolean {
+    // REVIEW: is there a better way to detect this?
+    if (bloomPage) return bloomPage.closest("#page-scaling-container") !== null;
+    return document.body.querySelector("div#page-scaling-container") !== null;
+}
+
 // Function to call to get everything ready for playing the game.
 // Things that get done here should usually be undone in undoPrepareActivity.
 export function prepareActivity(
@@ -64,6 +92,7 @@ export function prepareActivity(
     changePageAction: (next: boolean) => void,
 ) {
     currentPage = page;
+    adjustDraggablesForLanguage(page);
     currentChangePageAction = changePageAction;
     doShowAnswersInTargets(
         page.getAttribute("data-show-answers-in-targets") === "true",
@@ -442,7 +471,7 @@ function getVisibleEditables(container: HTMLElement) {
     return result;
 }
 
-function shuffle<T>(array: T[]): T[] {
+export function shuffle<T>(array: T[]): T[] {
     // review: something Copliot came up with. Is it guaranteed to be sufficiently different
     // from the correct answer?
     let currentIndex = array.length,
@@ -600,6 +629,13 @@ const getVisibleText = (elt: HTMLElement): string => {
 };
 
 const rightPosition = (draggableToCheck: HTMLElement, target: HTMLElement) => {
+    // Note: in some cases, some of the draggables and their targets are hidden (e.g.,
+    // the word spelling game keeps letter draggables that are not needed for the
+    // current word in case they are needed in another language.)
+    // These automatically register as correct, because offsetLeft and offsetTop return zero
+    // for display:none elements, so the position we get for the draggable and the target are
+    // both (0,0). This is what we want, since the user can't be faulted for not dragging
+    // invisible elements to the right place.
     const actualX = draggableToCheck.offsetLeft;
     const actualY = draggableToCheck.offsetTop;
     const correctX =
@@ -679,6 +715,9 @@ function showCorrectOrWrongItems(page: HTMLElement, correct: boolean) {
 
 function playSound(someElt: HTMLElement, soundFile: string) {
     const audio = new Audio(urlPrefix() + "/audio/" + soundFile);
+    if (IsRunningOnBloomDesktop()) {
+        audio.classList.add("bloom-ui"); // in case remove code fails, should make sure it doesn't get saved.
+    }
     audio.style.visibility = "hidden";
     // To my surprise, in BP storybook it works without adding the audio to any document.
     // But in Bloom proper, it does not. I think it is because this code is part of the toolbox,
--- a/src/event.ts
+++ b/src/event.ts
@@ -1,6 +1,5 @@
 // originally from http://stackoverflow.com/a/14657922/723299.
-// Temporarily duplicated in Bloom Player and Bloom Desktop (under Drag Activity),
-// but planned to become a Bloom Player artifact.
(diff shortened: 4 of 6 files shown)


andrew-polk (changes requested):
> Reviewed 4 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/357)* status: 4 of 6 files reviewed, 4 unresolved discussions (waiting on @StephenMcConnel)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISnhKN7h28Ne2Jl9dF:-OISnhKN7h28Ne2Jl9dG:b-trhjsr) (no related file):*
I added a couple comments, but there are enough review questions which I don't the answer to, I'm going to let @JohnThomson take a stab.
___
*[`src/scrolling.ts` line 18 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISnMYm7S5CKaVVT_jA:-OISnMYm7S5CKaVVT_jB:bhh6ezd) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/src/scrolling.ts#L18)):*
> ```XML
>     }
>     const isRunningOnBloomDesktop =
>         bloomPage.closest("#page-scaling-container") !== null;
> ```

Seems worth extracting this as a function since we use it in more than one place.
Either the function could accept a bloomPage or, I assume, we could use the body as you did in dragActivityRuntime.
___
*[`package.json` line 80 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISmHsB-o_WmnyqC1v_:-OISmHsB-o_WmnyqC1va:bdywvt3) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/package.json#L80)):*
> ```JSON
>         "dist/*.mp3",
>         "lib/*.js",
> 	"src/event.ts",
> ```

I know we can't have real comments here, but it seems worth adding one of our hack comments if possible.
Something like 
`"// files": "The four src/ files are directly exported so BloomDesktop can consume them as shared code.`
___
*[`package.json` line 80 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISmP8NECbhCd4FVaTl:-OISmP8NECbhCd4FVaTm:b-oru1kj) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/package.json#L80)):*
> ```JSON
>         "dist/*.mp3",
>         "lib/*.js",
> 	"src/event.ts",
> ```

These new lines have tab indentation instead of spaces.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/357)* status: 1 of 6 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`package.json` line 80 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISmHsB-o_WmnyqC1v_:-OIlmDtV4w-CZWx7EWOY:b-896fix) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/package.json#L80)):*

I know we can't have real comments here, but it seems worth adding one of our hack comments if possible.
Something like 
`"// files": "The four src/ files are directly exported so BloomDesktop can consume them as shared code.`
</blockquote></details>

Done.
___
*[`package.json` line 80 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISmP8NECbhCd4FVaTl:-OIlmEEv70dhP3nUAKH6:bt2vr1b) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/package.json#L80)):*

These new lines have tab indentation instead of spaces.
</blockquote></details>

Fixed.
___
*[`src/scrolling.ts` line 18 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISnMYm7S5CKaVVT_jA:-OIlodrP8cTFxK7guI2u:b-rl1j7p) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/src/scrolling.ts#L18)):*

Seems worth extracting this as a function since we use it in more than one place.
Either the function could accept a bloomPage or, I assume, we could use the body as you did in dragActivityRuntime.
</blockquote></details>

The reason i didn't use a single function in one place to begin with is, where does it live?  There isn't a natural home among these four files and I didn't want to create a new file.  I decided that dragActivityRuntime made as much sense as any of the files.  (which is where a function was written)

I moved the four files in BloomDesktop to a common shared/ folder, and adjusted the other BloomDesktop files accordingly to make this work.  This was more extensive for this changeset than I originally planned, but only the BloomDesktop specific files are critically important to review since the shared files are identical to those in bloom-player, apart from the "import $ from jquery" line in scrolling.ts.

andrew-polk (commented):
> Reviewed 3 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/357)* status: all files reviewed, 6 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/scrolling.ts` line 18 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISnMYm7S5CKaVVT_jA:-OJ0F7ZIAR-CATygG0nj:bbaehhg) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/src/scrolling.ts#L18)):*

The reason i didn't use a single function in one place to begin with is, where does it live?  There isn't a natural home among these four files and I didn't want to create a new file.  I decided that dragActivityRuntime made as much sense as any of the files.  (which is where a function was written)

I moved the four files in BloomDesktop to a common shared/ folder, and adjusted the other BloomDesktop files accordingly to make this work.  This was more extensive for this changeset than I originally planned, but only the BloomDesktop specific files are critically important to review since the shared files are identical to those in bloom-player, apart from the "import $ from jquery" line in scrolling.ts.
</blockquote></details>

Sorry; always feel free to push back if a suggestion doesn't seem worth the effort to make it work.

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/357)* status: all files reviewed, 6 unresolved discussions (waiting on @andrew-polk, @hatton, and @JohnThomson)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISnhKN7h28Ne2Jl9dF:-OJ0IbV7Dol8scvEYlmk:b6wnj2p) (no related file):*

I've done my best. I haven't duplicated the multi-way diff to try to confirm how the two versions should merge.

It's possible there should be more than four files. For example, I think I cheated and put some video-playback-related code in narration.ts rather than have yet another duplicate file to maintain. But now we have a clean way to handle it, there could be a sharedVideoFunctions.ts if we want, and maybe there's other stuff we should pull out.
Maybe that's too much to attempt in one pair of PRs.
</blockquote></details>

I'd rather stick with the four current files for now.
___
*[`src/dragActivityRuntime.ts` line 98 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OJ04u8dBCEhUrs-9E1w:-OJ0NEyk15k89Tl_QRXN:bhiu868) ([raw file](https://github.com/BloomBooks/bloom-player/blob/a3437c0623055a5a2abe6a5e02abf53e542a8711/src/dragActivityRuntime.ts#L98)):*

I think it is intended to be possible (but maybe not all the pieces are working yet?) to have a spelling game be multilingual. The translation of the word to spell may have a different number of letters, which requires a different number of draggables to be visible, as well as each showing the right letter for the word in that language. So if a book on BP switches language, this would be needed to update the visibility class on the draggables so only the non-empty ones show up. So my best guess is that it IS wanted (but may not be sufficient).
</blockquote></details>

Done.  (Removed test for IsRunningOnBloomDesktop)
___
*[`src/dragActivityRuntime.ts` line 157 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OJ05iu_5hUG6g6NYhQY:-OJ0NtWX-lYdrIffoPZP:b2mt8dd) ([raw file](https://github.com/BloomBooks/bloom-player/blob/a3437c0623055a5a2abe6a5e02abf53e542a8711/src/dragActivityRuntime.ts#L157)):*

@hatton do we want to hide video playback controls on overlay videos in Bloom desktop?
My guess: we definitely do while the game is in play mode, we want that to be as much like BP as possible. Probably simplest to just always suppress them. So, unless JohnH disagrees, I'd say just remove the test for IsRunningOnBloomDesktop.
Then again, maybe this code is obsolete altogether? I think we did work not very long ago to stop using the built-in controls on videos, and just have play, pause, and/or restart buttons. I would guess we want those (and not the standard playback controls) on all videos in both programs.
</blockquote></details>

Done.  (removed the test for !IsRunningOnBloomDesktop
___
*[`src/dragActivityRuntime.ts` line 724 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OJ06MZy8nMKfzObauX-:-OJ0Of-f6zFR1ZCPpY5X:b-40taqu) ([raw file](https://github.com/BloomBooks/bloom-player/blob/a3437c0623055a5a2abe6a5e02abf53e542a8711/src/dragActivityRuntime.ts#L724)):*

No. It's only use it to stop things being persisted, which BP never does.
Conversely, it's harmless, so I'm not sure the complication of an "if" to prevent it is worthwhile.
</blockquote></details>

Done.  I removed the comment, not the test.
___
*[`src/event.ts` line 2 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OJ06np8AKexpiTch2Ww:-OJ0Ozfw-kCs__ujgq2N:b-c58rbn) ([raw file](https://github.com/BloomBooks/bloom-player/blob/a3437c0623055a5a2abe6a5e02abf53e542a8711/src/event.ts#L2)):*

Is artifact the right word? I think I may have been thinking of TeamCity, where build outputs are called that, but I'm not sure it's the right term for npm. The point is that as well as being used by BP, it is copied to the dist directory to be available for use by clients like BE.
</blockquote></details>

I changed "artifact" to "source file" in the comment
___
*[`src/scrolling.ts` line 18 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OISnMYm7S5CKaVVT_jA:-OJ0P94e-lB05AcefFMX:b-5oysoi) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7ca005e86167c598046e645f93c7f753e8fa852c/src/scrolling.ts#L18)):*

Sorry; always feel free to push back if a suggestion doesn't seem worth the effort to make it work.
</blockquote></details>

I think moving the four files into the same subfolder in Bloom  Desktop was worthwhile in its own right.
___
*[`src/scrolling.ts` line 299 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-OJ09NsQ4C5dGcEwC0Pu:-OJ0Q3cV1pSqS9epzuzO:b-97ays3) ([raw file](https://github.com/BloomBooks/bloom-player/blob/a3437c0623055a5a2abe6a5e02abf53e542a8711/src/scrolling.ts#L299)):*

Hard to know how far off 1 would make no difference, except by computing the adjustment and seeing if it is less than a pixel. And even "less than a pixel" might be cumulative with other fractional pixels. I'm inclined to leave it as is...or even take out the special case for "1" if it is not common enough to be much use.
</blockquote></details>

Not having any scaling is fairly common in BloomDesktop.  It may be less so in bloom-player environments.
I removed the comment asking the question.  :-)

JohnThomson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/357#-:-OJ4Lk7G6E4JqpqkZIl4:bnfp4nl)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

andrew-polk (approved):

PR #366, opened 2025-03-31
fix: Ensure that shuffled letters in words look shuffled (BL-14446)

StephenMcConnel:
This is the bloom-player side of the fix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/366)

### DIFF
--- a/src/shared/dragActivityRuntime.ts
+++ b/src/shared/dragActivityRuntime.ts
@@ -482,20 +482,74 @@ function getVisibleEditables(container: HTMLElement) {
     return result;
 }
 
-export function shuffle<T>(array: T[]): T[] {
-    // review: something Copilot came up with. Is it guaranteed to be sufficiently different
-    // from the correct answer?
-    let currentIndex = array.length,
-        randomIndex;
-    while (0 !== currentIndex) {
-        randomIndex = Math.floor(Math.random() * currentIndex);
-        currentIndex--;
-        [array[currentIndex], array[randomIndex]] = [
-            array[randomIndex],
-            array[currentIndex],
-        ];
+export function shuffle<T>(
+    array: T[],
+    areItemsTheSame?: (a: T, b: T) => boolean,
+): T[] {
+    // something Copilot came up with.  Otherwise known as the Fisher–Yates Shuffle as discussed
+    // at https://stackoverflow.com/questions/2450954/how-to-randomize-shuffle-a-javascript-array.
+    // Getting the output to always be visibly different when input items can visually be the same (as for
+    // letters in a word that are repeated) adds complications before and after the guts of the computation.
+    // See BL-14446.
+
+    // simplify code below by providing a default function
+    if (!areItemsTheSame)
+        areItemsTheSame = (a: T, b: T) => {
+            return a === b;
+        };
+    let allItemsAreTheSame = true;
+    for (let i = 1; i < array.length; ++i) {
+        if (!areItemsTheSame(array[0], array[i])) {
+            allItemsAreTheSame = false;
+            break;
+        }
     }
+    if (allItemsAreTheSame) return array; // shuffling not needed, wouldn't work anyway
+
+    const originalArray = array.slice();
+    let bestChanges = 0;
+    let bestResult = originalArray;
+    let loopCount = 0;
+    // Note that each time through the loop uses the output of the previous iteration
+    // as its input.
+    while (loopCount++ < 10) {
+        // BEGIN Fisher-Yates shuffle
+        let currentIndex = array.length,
+            randomIndex;
+        while (0 !== currentIndex) {
+            randomIndex = Math.floor(Math.random() * currentIndex);
+            currentIndex--;
+            [array[currentIndex], array[randomIndex]] = [
+                array[randomIndex],
+                array[currentIndex],
+            ];
+        }
+        // END Fisher-Yates shuffle
+
+        // Check whether the shuffling has had an actual visible effect.
+        // Save the array with the largest number of changes while the
+        // loop is processing.
+        let changes = 0;
+        for (let i = 0; i < array.length; ++i) {
+            if (!areItemsTheSame(array[i], originalArray[i])) ++changes;
+        }
+        if (changes === array.length) return array; // It can't get better on this simplistic measure.
+        if (changes > bestChanges) {
+            bestChanges = changes;
+            bestResult = array.slice();
+        }
+    }
+    // callers may depend on the array being modified in place.
+    for (let i = 0; i < array.length; ++i) array[i] = bestResult[i];
     return array;
+}
+
+// This method can be passed to shuffle to test letters embedded in draggable HTML elements.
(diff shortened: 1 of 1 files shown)


hatton (commented):
The diff is making it look like you had to add a lot of code. I'm surprised you can't just use whatever was there and re-run when you detect it's unscrambled?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-:-OMiUTPa2GCG0-FQKgMC:bbkkw5r)* status: 0 of 1 files reviewed, all discussions resolved

StephenMcConnel (commented):
Apart from the prologue where I detect an (improbable but not impossible) word entered like "ooo" where all the letters are the same, that's essentially what I did.  The problem is that detecting that the two e's in feet are interchangeable from the human's perspective is somewhat nontrivial.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-:-OMiWkkH3o5Af3n8R0AK:b-uxu660)* status: 0 of 1 files reviewed, all discussions resolved

JohnThomson (changes requested):
I don't see a change anywhere to call the improved shuffle with the new function. Maybe that's on the BE side?

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-:-OMmpOs60lVBXwWhJomF:brbv7zr)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`src/shared/dragActivityRuntime.ts` line 494 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-OMmo70745zu77ymK23t:-OMmo70745zu77ymK23u:bmxfhtv) ([raw file](https://github.com/BloomBooks/bloom-player/blob/9a8bb208a90e67709e70ab35c35bd1952be16de9/src/shared/dragActivityRuntime.ts#L494)):*
> ```XML
>     let allItemsAreTheSame = true;
>     for (let i = 1; i < array.length; ++i) {
>         if (areItemsTheSame) {
> ```

could we get rid of this and the outer "if" in both loops by (outside the loop) saying areItemsTheSame = areItemsTheSame ?? (a,b) => a === b? Or even (not sure if JS allows this) providing that function as a default value for the argument?
___
*[`src/shared/dragActivityRuntime.ts` line 513 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-OMmqTvS16DLcl1jR8s-:-OMmqTvS16DLcl1jR8s0:bz6gilc) ([raw file](https://github.com/BloomBooks/bloom-player/blob/9a8bb208a90e67709e70ab35c35bd1952be16de9/src/shared/dragActivityRuntime.ts#L513)):*
> ```XML
>     // In testing, I saw the loop count grow as high as 4 for some
>     // words with repeated letters.
>     while (changes === 0 && loopCount++ < 10) {
> ```

Feels like we could afford to try a bit harder. This is not a very expensive loop and the whole thing happens once per keystroke. The resulting dom manipulations probably take a lot longer.
I would think the ideal is something like changes ===array.length, though for very short words knowing that every letter is in the wrong place might help too much. In fact, I wish we could evaluate a bit better how far the letter is from its expected place.
A start would be to always try 10 times and keep the result with the highest change score. (Maybe stop if you get a perfect array.length.)
Another possible improvement would be to evaluate "close", where close counts how often areItemsTheSame(originalArray[i], array[(i+1) mod array.length]) is true. Shuffles with equal 'changes' scores could be distinguished by their 'close' scores. Or just generate a single value like score = changes + close/10 (or maybe close is more of a concern than that...maybe close/5?)

hatton:
A chatgpt query to help think about this:

"i need an algorithm for shuffling letters of a word for a literacy game. Obviously a shuffle that leads to something that looks like the original is bad. Shuffles that only require one move aren't as good as ones that require many moves."

StephenMcConnel (commented):
Yes, that's on the BloomDesktop side.  The local call feeds in an array of strings, which I think works okay with the default equality test.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-:-OMn9QcV1YBm4B6CqmX7:bcmm4kk)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/shared/dragActivityRuntime.ts` line 494 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-OMmo70745zu77ymK23t:-OMnI_oE0yd2Mja6FTo7:b-gjien3) ([raw file](https://github.com/BloomBooks/bloom-player/blob/9a8bb208a90e67709e70ab35c35bd1952be16de9/src/shared/dragActivityRuntime.ts#L494)):*

could we get rid of this and the outer "if" in both loops by (outside the loop) saying areItemsTheSame = areItemsTheSame ?? (a,b) => a === b? Or even (not sure if JS allows this) providing that function as a default value for the argument?
</blockquote></details>

Your syntax didn't seem to work, but I implemented the idea.
___
*[`src/shared/dragActivityRuntime.ts` line 513 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/366#-OMmqTvS16DLcl1jR8s-:-OMnJ-AH8FvveMvACaHt:brbc7la) ([raw file](https://github.com/BloomBooks/bloom-player/blob/9a8bb208a90e67709e70ab35c35bd1952be16de9/src/shared/dragActivityRuntime.ts#L513)):*

Feels like we could afford to try a bit harder. This is not a very expensive loop and the whole thing happens once per keystroke. The resulting dom manipulations probably take a lot longer.
I would think the ideal is something like changes ===array.length, though for very short words knowing that every letter is in the wrong place might help too much. In fact, I wish we could evaluate a bit better how far the letter is from its expected place.
A start would be to always try 10 times and keep the result with the highest change score. (Maybe stop if you get a perfect array.length.)
Another possible improvement would be to evaluate "close", where close counts how often areItemsTheSame(originalArray[i], array[(i+1) mod array.length]) is true. Shuffles with equal 'changes' scores could be distinguished by their 'close' scores. Or just generate a single value like score = changes + close/10 (or maybe close is more of a concern than that...maybe close/5?)
</blockquote></details>

Any measurement based purely on number of changes is not always going to work well.  Consider this: "elephant" being shuffled to "phantele" has a perfect score for the number of changes, but is hardly shuffled the way we would want.  But measuring the maximum common substring length seems more complex than is worthwhile in practice.
I've taken your suggestion to always run the loop 10 times, and return the shuffled array that has the largest number of changes, stopping earlier at the first perfect score if one occurs.
Measuring the difference could become arbitrarily complex.  I think what we have will probably work well enough in practice.  The documentation could point out that if someone isn't happy with the scrambling, they just need to re-enter the same word and a different scramble is almost certain to occur.

JohnThomson (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/366)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #375, opened 2025-05-14
fix: Hide answers from drag activities when changing language (BL-14776)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/375)

### DIFF
--- a/src/activities/activityManager.ts
+++ b/src/activities/activityManager.ts
@@ -230,6 +230,16 @@ export class ActivityManager {
                 "prepared",
             );
         }
+        // It would be nice if this next bit of work could be done in dragActivityRuntime.prepareActivity(),
+        // but that is called after the page is already in the DOM and the answer has been (briefly) visible
+        // to the user. So we do it here, before the page is shown at all. (BL-14776)
+        const visibleTargets = pageDiv.querySelectorAll(
+            "[data-target-of] .bloom-visibility-code-on",
+        );
+        for (let i = 0; i < visibleTargets.length; i++) {
+            const element = visibleTargets[i] as HTMLElement;
+            element.classList.remove("bloom-visibility-code-on");
+        }
     }
 
     public doInitialSoundAndAnimation() {


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/375)* status: all files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

___
*[`src/activities/activityManager.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/375#-OQFelIt5GyZm4O2k29u:-OQFelIt5GyZm4O2k29v:b-viyqv7) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7e810efe46fe29821fae7010b9fa030e910b7896/src/activities/activityManager.ts#L239)):*
> ```XML
>             const element = visibleTargets[i] as HTMLElement;
>             element.classList.remove("bloom-visibility-code-on");
>         }
> ```

Can we move this to dragActivityRuntime's prepareActivity?

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/375)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/activities/activityManager.ts` line 239 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/375#-OQFelIt5GyZm4O2k29u:-OQFoXK70A5DD3L3A4Rw:bcdlfx3) ([raw file](https://github.com/BloomBooks/bloom-player/blob/7e810efe46fe29821fae7010b9fa030e910b7896/src/activities/activityManager.ts#L239)):*

Can we move this to dragActivityRuntime's prepareActivity?
</blockquote></details>

No, that gets called too late in the process of reloading the page with the new language.  I spent a couple of hours trying to get something working in that method to no avail.
I added a comment explaining why this code is here.

andrew-polk (commented):
I'm exploring some other options.
I think we have a fundamental problem that the target even has content when we save the page.
If we fix that, it will also solve the problem that the PDF is currently displaying the targets but shouldn't. But I also don't know how we would show targets before generating the PDF in the case where we want to (user has turned on that option).

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/375#-:-OQKNym44z90mXdiMzQh:bm1fo0)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @StephenMcConnel)

andrew-polk:
Superceded by https://github.com/BloomBooks/BloomDesktop/pull/7092

PR #376, opened 2025-05-20
fix: Prevent missing audio from wiping out game header styling (BL-14797)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/376)

### DIFF
--- a/src/shared/narration.ts
+++ b/src/shared/narration.ts
@@ -752,13 +752,13 @@ function setSoundAndHighlight(
     disableHighlightIfNoAudio: boolean,
     oldElement?: Element | null | undefined,
 ) {
+    setSoundFrom(newElement);
     setHighlightTo({
         newElement,
         shouldScrollToElement: true, // Always true in bloom-player version
         disableHighlightIfNoAudio,
         oldElement,
     });
-    setSoundFrom(newElement);
 }
 
 function setHighlightTo({
@@ -785,10 +785,13 @@ function setHighlightTo({
 
     removeAudioCurrent((oldElement || newElement) as HTMLElement);
 
-    if (disableHighlightIfNoAudio) {
-        const mediaPlayer = getPlayer();
+    const mediaPlayer = getPlayer();
+    // The "error" may just be that audio has not been recorded for this element.
+    // If the audio is missing, then strange things happen in the interaction between
+    // narration and drag activity text.  See BL-14797
+    const hasError = mediaPlayer.error !== null;
+    if (!hasError && disableHighlightIfNoAudio) {
         const isAlreadyPlaying = mediaPlayer.currentTime > 0;
-
         // If it's already playing, no need to disable (Especially in the Soft Split case, where only one file is playing but multiple sentences need to be highlighted).
         if (!isAlreadyPlaying) {
             // Start off in a highlight-disabled state so we don't display any momentary highlight for cases where there is no audio for this element.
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/376)* status: all files reviewed, 2 unresolved discussions (waiting on @StephenMcConnel)

___
*[`src/shared/narration.ts` line 792 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/376#-OQsRImi7JhBt-W2FVfk:-OQsR1gn3OFLhSbVJBcB:b8kvlfg) ([raw file](https://github.com/BloomBooks/bloom-player/blob/686bbf95895e74bc9aa9b72e3d0e964fb95fb2cf/src/shared/narration.ts#L792)):*
> ```XML
>     // If the audio is missing, then strange things happen in the interaction between
>     // narration and drag activity text.  See BL-14797
>     const hasError = mediaPlayer.error !== null;
> ```

How do we know that the error is caused by a problem on the current element rather than a previous one? For example, will this interfere with playing the recording for the second element on the page, if the first one doesn't have a recording?
Note that setSoundAndHighlight calls this method BEFORE it updates the src on the player, which is what resets the error field. (Maybe that could just be reversed...)
___
*[`src/shared/narration.ts` line 808 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/376#-OQsRfeg5V7itkO-Xt0m:-OQsRfeg5V7itkO-Xt0n:b-fr3lh) ([raw file](https://github.com/BloomBooks/bloom-player/blob/686bbf95895e74bc9aa9b72e3d0e964fb95fb2cf/src/shared/narration.ts#L808)):*
> ```XML
>             mediaPlayer.addEventListener("error", () => {
>                 newElement.classList.remove(kSuppressHighlightClass);
>                 newElement.classList.remove("ui-audioCurrent");
> ```

Since the purpose of suppressHighlight is to prevent the highlight flickering when audioCurrent is applied but turns out not to be correct, it would be more natural to remove audioCurrent first (though it probably doesn't matter since I don't expect a re-render between these two lines).

StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/376)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @JohnThomson)

___
*[`src/shared/narration.ts` line 792 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/376#-OQsRImi7JhBt-W2FVfk:-OQtZLIyAQmxX-X2H0tZ:biqgy7n) ([raw file](https://github.com/BloomBooks/bloom-player/blob/686bbf95895e74bc9aa9b72e3d0e964fb95fb2cf/src/shared/narration.ts#L792)):*

How do we know that the error is caused by a problem on the current element rather than a previous one? For example, will this interfere with playing the recording for the second element on the page, if the first one doesn't have a recording?
Note that setSoundAndHighlight calls this method BEFORE it updates the src on the player, which is what resets the error field. (Maybe that could just be reversed...)
</blockquote></details>

The code was already assuming the proper player state by testing player.currentTime > 0.  In testing, this doesn't seem to interfere with playing the recording for the second element when the first one doesn't have a recording.  But I reversed the order of setSoundFrom() and setHighlightTo() in the calling method.
___
*[`src/shared/narration.ts` line 808 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/376#-OQsRfeg5V7itkO-Xt0m:-OQtXv_p42sKWhCVdJiU:b-896fix) ([raw file](https://github.com/BloomBooks/bloom-player/blob/686bbf95895e74bc9aa9b72e3d0e964fb95fb2cf/src/shared/narration.ts#L808)):*

Since the purpose of suppressHighlight is to prevent the highlight flickering when audioCurrent is applied but turns out not to be correct, it would be more natural to remove audioCurrent first (though it probably doesn't matter since I don't expect a re-render between these two lines).
</blockquote></details>

Done.

PR #386, opened 2025-08-05
BL-15020: Ken Burns effect now clones entire canvas

NJStriebel:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/386)


JohnThomson (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/386)* status: all files reviewed, 20 unresolved discussions (waiting on @NJStriebel)

___
*[`src/shared/animation.ts` line 11 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OWzzHd993Xm4CYevMzh:-OWzzHd993Xm4CYevMzi:b-cowh32) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L11)):*
> ```XML
> // Enhance: Jeffrey has some code from one of the hackathon students that
> // makes the animation go at a more even pace
> // Enhance: I think it might be possible to do the animation more simply with CSS transitions.
> ```

I think those enhance comments are obsolete? At least the last one.
___
*[`src/shared/animation.ts` line 26 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OWzzj6N0P3PFoLQZu8Y:-OWzzj6N0P3PFoLQZu8Z:b-54gtrd) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L26)):*
> ```XML
>     }
> 
>     // Get the animatable clone of the animatable image container, if we already made it.
> ```

Since this file got moved, you get to take responsibility for all of it :-).
Most places in this file, the name 'image container' is obsolete. For current books, the element you want to animate is a bloom-canvas. (We do still have image containers, but only in image overlays (canvas elements). (Well, in bloom editor, the background image is inside a special canvas element and therefore inside an image container.)
Unfortunately, bloom player may still encounter books that haven't had this migration, so the thing we now call a bloom-canvas may actually have the class bloom-imageContainer instead of bloom-canvas. But even then, conceptually we're getting a bloom-canvas...a top-level thing that typically has one background image, and possibly some overlays.
Could you fix wherever the name 'image container' is used to mean the bloom-canvas?
___
*[`src/shared/animation.ts` line 50 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-0IS630dWpJRfC_ki:-OX-0IS630dWpJRfC_kj:bqaatrs) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L50)):*
> ```XML
> 
>     // Search for an image container that has the properties we need for animation.
>     public static getAnimatableImageContainer(page: HTMLElement): HTMLElement {
> ```

Method names should be fixed, too...this is really getAnimatableBloomCanvas, even though in obsolete BloomPubs it may actually return an image container.
___
*[`src/shared/animation.ts` line 64 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-Hxkb16sHwDzImDAu:-OX-Hxkb16sHwDzImDAv:b-oenhbc) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L64)):*
> ```XML
>     }
> 
>     private animationEngines : { [key:string]:TransformBasedAnimator } = {}; //we keep the animation engines in a map of "pageNumber":engine to avoid making a new one each time we load a page
> ```

Say a bit more about why you want to avoid that (or consider not doing so). The constructor of TransformBasedAnimator does not look very time-consuming, and keeping this cache costs space and complexity. Maybe it lets you hang on to any work you did loading the image to get dimensions? But that might be obsolete with the work you're doing to pre-calculate them.
___
*[`src/shared/animation.ts` line 65 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-9w1MB0IAVg2LvSRD:-OX-9w1MB0IAVg2LvSRE:brrftdv) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L65)):*
> ```XML
> 
>     private animationEngines : { [key:string]:TransformBasedAnimator } = {}; //we keep the animation engines in a map of "pageNumber":engine to avoid making a new one each time we load a page
>     public PlayAnimations: boolean; // true if animation should occur
> ```

Let's say a little more, here. There are various reasons for animation not to occur, at least
- page doesn't have anything animatable
- in portrait orientation
- paused
Does this track all three (and any others), or just some of them?
___
*[`src/shared/animation.ts` line 82 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-0gDz2sjgbq9_wU9B:-OX-0gDz2sjgbq9_wU9C:b-p4lqam) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L82)):*
> ```XML
> 
>     public HandlePageBeforeVisible(page: HTMLElement) {
>         if (!this.shouldAnimate(page)){      
> ```

Looks like we have trailing spaces on this line...I would have expected Prettier to prevent this. Also in other places.
___
*[`src/shared/animation.ts` line 118 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-1MBuDfhbXb9t07JJ:-OX-1MBuDfhbXb9t07JK:b6t9vu6) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L118)):*
> ```XML
>     // May be called when we are not paused; should do nothing if so.
>     public PlayAnimation(page:HTMLElement) {
>         const engine = this.animationEngines[page.getAttribute("data-page-number")];
> ```

Using data-page-number worries me a little. Some pages in a book don't have one. (One day we might support animating the front cover, for example.) Could we just use the page itself as a key? Or do those objects keep getting destroyed and recreated? If it's not safe to use the page as a key, are we sure that the (different) element that now represents that page will be properly initialized for animation, or can we count on the engine to do that if necessary?
(Or, as suggested above, maybe it's not worth keeping more than the current one around?)
___
*[`src/shared/animation.ts` line 170 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-2aZv84p-C2vO31NZ:-OX-2aZv84p-C2vO31N_:bskf5fw) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L170)):*
> ```XML
>         //          </div>
>         //      </div>
>         // and insert it into the bloom-page div as the first element.
> ```

Would you check that the above is actually consistent with the current state of things? I think at least bloom-imageContainer should as usual change to bloom-canvas, but there may be other inconsistencies.
___
*[`src/shared/animation.ts` line 172 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-59GN85r2-tOOFDrL:-OX-59GN85r2-tOOFDrM:b3n6bq4) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L172)):*
> ```XML
>         // and insert it into the bloom-page div as the first element.
> 
>         let hidePageDiv = page.firstElementChild as HTMLDivElement;
> ```

Every time I read this code, I'm less happy with (my) names hidePageDiv, wrapDiv, and animationDiv. Can you think of any improvement? IIRC, they are something like "rootOfAnimationThatHidesPage", "fixedDivThatClipsTheMovingOne" and "cloneOfBloomCanvasThatMoves". Those might be a bit long, but they would also convey more. (If you change them, please update the classes that identify them to match.)
___
*[`src/shared/animation.ts` line 200 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-3bokEusSlneepMk6:-OX-3bol1bPB8iDuCN1r:b-pjdkq8) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L200)):*
> ```XML
>             if(!animationView.hasAttribute("data-imgsizebasedon")){
>                 requestAnimationFrame(checkForBackgroundImage);
>             }
> ```

I'm not getting the block above. It doesn't seem to have anything to do with loading the image: the clone either has a backgroundImage in its style or does not, and that won't change after we get done cloning it, unless something is running in the background to do so. It's not as if (here) we're waiting for the image to load so we can get its dimensions. I'm wondering whether this function did more (and the comment applied) in the past, and now we can remove it? Or maybe the background image gets filled in when Swiper does its lazy loading? If so, deserves comment (or as we discussed this morning, we may want to just not use Swiper lazy image loading).
___
*[`src/shared/animation.ts` line 234 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-6OGy7PjyFbpA7ckK:-OX-6OGy7PjyFbpA7ckL:b2yziyf) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L234)):*
> ```XML
>         const animationView = Animation.getAnimationView(page);
> 
>         //stop the old engine's animation so it can's compete with the new animation
> ```

can's -> can't
___
*[`src/shared/animation.ts` line 273 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-7bPn3kcidhrBr0bh:-OX-7bPn3kcidhrBr0bi:b-monjx5) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L273)):*
> ```XML
>                 }
>                 else{
>                     requestAnimationFrame(waitForImageAndSetAspectRatio);
> ```

Again, it's not obvious to me what would cause it to get a style.backgroundImage if it doesn't have one originally. Maybe this is a consequence of swiper lazy loading? Worth a comment if so.
___
*[`src/shared/animation.ts` line 326 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-8nf26ojsC4njQ4Pf:-OX-8nf26ojsC4njQ4Pg:b-k08ryd) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L326)):*
> ```XML
>             //our top left corner moved half the difference between the original canvas's dimensions and the wrap div's dimensions
>             canvas.style.top = `${(wrapDivDimensions[1] - originalDimensions[1])/2}px`;
>             canvas.style.left = `${(wrapDivDimensions[0] - originalDimensions[0])/2}px`;
> ```

I think setting canvas.style.transformOrigin = "top left" might remove the need to do this, though it may cause complications elsewhere. Only if it simplifies things overall!
___
*[`src/shared/animation.ts` line 382 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-Aflq3PJxK0kFec_6:-OX-Aflr9A14rXujk-nh:b942r7z) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L382)):*
> ```XML
>     }
> 
>     public shouldAnimate(page: HTMLElement): boolean {
> ```

As with PlayAnimations, it would help to comment here exactly what set of factors that affect whether we shouldAnimate this takes into account. (Clearly it adds orientation!)
___
*[`src/shared/animation.ts` line 405 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-BznHECAhPxtIGIld:-OX-BznHECAhPxtIGIle:b-ploy32) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L405)):*
> ```XML
>     private lastFrameTime : number;
>     private totalElapsedTime : number;
>     private duration : number;
> ```

comment: milliseconds
___
*[`src/shared/animation.ts` line 410 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-BKLK1gRD0MmOAcsC:-OX-BKLK1gRD0MmOAcsD:bnuulsi) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L410)):*
> ```XML
>     
>     //initial and final rect strings come in as space-separated "left top width height"
>         //all four parameters are reperesented as fractions of the canvas
> ```

canvas = toAnimate? Maybe say ...fractions of the toAnimate element's dimensions (typically a clone of a bloom-canvas, adjusted to be its original size)
___
*[`src/shared/animation.ts` line 414 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-BnKZDtZcx3g2MWPW:-OX-BnKZDtZcx3g2MWPX:b-5fu0sg) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L414)):*
> ```XML
>     //toAnimate is the whole canvas containing the images to be animated.
>         //note that toAnimate needs to be the child of an element that has the correct aspect ratio, has overflow set to hidden, AND has a transform applied
>             //the parent's transform can just be translateZ(0), but it needs to be moved out of the page stacking context in the same way as its child in order for overflow:hidden to have any effect
> ```

Something funny happening with indentation? Prettier should clean this sort of thing up.
___
*[`src/shared/animation.ts` line 449 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-F8BmAi6Gt5GiCOCJ:-OX-F8BmAi6Gt5GiCOCK:b8q9m80) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L449)):*
> ```XML
>             if(this.finalScale == 1) this.finalScale = 1.0001;
> 
>             //the amount we need to move the top left corner relative to the center is 50% - (1/2 width + position)
> ```

Again, this might be simpler using transformOrigin top left...I haven't tried to work it out.
___
*[`src/shared/animation.ts` line 503 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-G3S89WrWD5UpCuUs:-OX-G3S89WrWD5UpCuUt:b-r44jyx) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L503)):*
> ```XML
> 
>     public play(){
>         this.paused = false;
> ```

Maybe marginally better to call advanceAnimation here, and NOT have advanceAnimation request another animation frame when it is paused? The busy-waiting of requesting animation frames when not doing anything will tend to use power on a mobile device.
(But: I see that advanceAnimation is updating lastFrameTime, so that might get tricky. Don't worry about this if it seems to complicate things much.)
___
*[`src/shared/animation.ts` line 507 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-H21kA8DVKQfbCCgm:-OX-H21kA8DVKQfbCCgn:b-7is03s) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L507)):*
> ```XML
> 
>     public startAnimation(){
>         this.totalElapsedTime = 0;
> ```

Should we also set this.lastFrameTime here, in case any significant time has passed since construction?

JohnThomson (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/386)* status: 4 of 5 files reviewed, 5 unresolved discussions (waiting on @NJStriebel)

___
*[`src/shared/animation.ts` line 26 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OWzzj6N0P3PFoLQZu8Y:-OX0HiOJ-fGX3AjgmR08:b-p6d33m) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L26)):*
<details><summary><i>Previously, NJStriebel (Nathan Striebel) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

You've done a lot, but I think you missed the comment right here :-)
___
*[`src/shared/animation.ts` line 64 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-Hxkb16sHwDzImDAu:-OX0PpOh2evWNwHTynBM:bgpd5z2) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L64)):*
<details><summary><i>Previously, NJStriebel (Nathan Striebel) wrote…</i></summary><blockquote>

Done. This approach solved a few problems related to pausing.

I initially did this in response to a bug where I was sometimes making two animation engines for the same page. In that case, they were performing in parallel, setting the same transform at the same time. That was fine until you try to pause. Then one engine would pause and the other would keep moving the canvas.

I found the place where I was making two calls to the method that makes the animation engine and removed that bug, but the fact that I had that problem revealed another potential problem:

Let's say you open page 1, switch to page 2, then quickly switch back to page 1. The desired behavior is that page 1 would start its animation over from the beginning. If we create a new animation engine without destroying the old animation engine, we could have both copies attempting to modify the transform of the same div.
</blockquote></details>

So what we're really wanting to ensure is that only one animation is ever running. Wouldn't it be enough to keep track of the current one, and if we are about to make a new one for whatever reason and there is a current one, stop it?
Not sure that's enough simpler to be worth changing, though, given that you have this working. Moreover, if we achieve fading from one page to the next, we may want two running at the same time. (Though I don't think so...we probably don't want to start fading to the next page until the current page is all the way finished?) 
I'm seeing a stronger reason, though, to possibly avoid this map. TranslationEngine has a reference to the page it is animating. As long as this essentially global map has a reference to the TE, any page we've animated can't be garbage collected, even if it's been removed from the DOM.
Motion books probably aren't long enough for this to be a big problem. It's pages with videos that get really expensive. So if there's not time to change this, I'm OK with merging as-is. We can always optimize that later.
___
*[`src/shared/animation.ts` line 82 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-0gDz2sjgbq9_wU9B:-OX0IZAT-mZFfXd2CeHO:bhj44it) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L82)):*
<details><summary><i>Previously, NJStriebel (Nathan Striebel) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I'm still seeing a row of little red dots at the end of the line.
___
*[`src/shared/animation.ts` line 118 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-1MBuDfhbXb9t07JJ:-OX0IjXU8syS2zRbWTvk:b-ovwaw7) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L118)):*
<details><summary><i>Previously, NJStriebel (Nathan Striebel) wrote…</i></summary><blockquote>

Done.

How do you feel about using page.id? As far as I can tell, that stays consistent as swiper destroys/reloads the page.
</blockquote></details>

Good idea.
___
*[`src/shared/animation.ts` line 326 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-8nf26ojsC4njQ4Pf:-OX0ME-o12d-zAtJeFFF:b-wkh1t3) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L326)):*
<details><summary><i>Previously, NJStriebel (Nathan Striebel) wrote…</i></summary><blockquote>

Done.

I tried it, and it seemed to keep the canvas centered in the wrapper, which is what these two lines were meant to fix. From the documentation I read, transform-origin moves the element, applies the transformation, then moves it back to its original position. We don't want it moved back to its original position.
</blockquote></details>

That's one way of visualizing it. Another (that I find more helpful) is to think of it as, you are specifying the point that doesn't change while you do whatever it is. So the example in the doc is, it rotates around that point. Likewise, if you scale with a top left transform origin, the top left of the canvas shouldn't move. With no transform, I would expect the top lefts to be aligned, since the canvas is the only child of the wrapper.
But you have something that works, and changing it is not helping, so let's leave it.
___
*[`src/shared/animation.ts` line 382 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-Aflq3PJxK0kFec_6:-OX0NNkg-I8nwmW-baSO:b-dky0r5) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L382)):*
<details><summary><i>Previously, NJStriebel (Nathan Striebel) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Is it worth adding ...but doesn't care whether animation is paused?
___
*[`src/shared/animation.ts` line 414 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-BnKZDtZcx3g2MWPW:-OX0Nrp04taD5JS1PoDR:bddkm5x) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L414)):*
<details><summary><i>Previously, NJStriebel (Nathan Striebel) wrote…</i></summary><blockquote>

Done. That was me. I indented it in the way that I thought organzied it best.
</blockquote></details>

Would you get Andrew (who is better at this sort of thing) to help you check that you have Prettier configured in our standard way? I'm pretty sure it will then remove your extra indentation. For that reason, and because I find it clearer, I usually put indentation that is meant to organize the structure of a comment inside the comment, like 
```
// canvasToAnimate...
//     note that...
//     the parent's...
```
___
*[`src/shared/animation.ts` line 513 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX0OxuJ9cNhMw4BfRCa:-OX0OxuK0qQTO3mHpDLD:bemsmnj) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L513)):*
> ```XML
> 
>     public endAnimation(){
>         this.paused = true;
> ```

I think not setting paused means that advanceAnimation will run one more time and jump it to the 'final' state. Is that what we want? Or don't we care? Maybe worth a comment.

andrew-polk:
*[`src/shared/animation.ts` line 414 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-BnKZDtZcx3g2MWPW:-OX0UCoME9VtRVbkGGte:b-a3pd78) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L414)):*

Would you get Andrew (who is better at this sort of thing) to help you check that you have Prettier configured in our standard way? I'm pretty sure it will then remove your extra indentation. For that reason, and because I find it clearer, I usually put indentation that is meant to organize the structure of a comment inside the comment, like 
```
// canvasToAnimate...
//     note that...
//     the parent's...
```
</blockquote></details>

Indeed, this file has not been formatted by prettier.
When I open it on your branch/fork and save, I get a lot of formatting changes. 
: (

andrew-polk:
*[`src/shared/animation.ts` line 414 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX-BnKZDtZcx3g2MWPW:-OX0U_RQ4NWwxPgYC3qe:b-ef8vs3) ([raw file](https://github.com/BloomBooks/bloom-player/blob/39e8e981099a2c9b0853534f021a85415e08b4bf/src/shared/animation.ts#L414)):*

Indeed, this file has not been formatted by prettier.
When I open it on your branch/fork and save, I get a lot of formatting changes. 
: (
</blockquote></details>

And yes, prettier will destroy your intentional indentation here.
You'll need to do something like 
```
// initial and final rect strings come in as space-separated "left top width height"
//    all four parameters are reperesented as fractions of the canvas
// duration is the length of the animation in seconds
// toAnimate is the whole canvas containing the images to be animated.
//    note that toAnimate needs to be the child of an element that has the correct aspect ratio, has overflow set to hidden, AND has a transform applied
//        the parent's transform can just be translateZ(0), but it needs to be moved out of the page stacking context in the same way as its child in order for overflow:hidden to have any effect

```

By the way, you can do Ctrl-Shift-P, Format document, to make it format.

JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/386)* status: 3 of 5 files reviewed, 4 unresolved discussions (waiting on @NJStriebel)

___
*[`src/shared/animation.ts` line 66 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9PY7w-80Geu0Kf2fH:-OX9PY7w-80Geu0Kf2fI:bpfjj65) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L66)):*
> ```XML
>     }
> 
>     //we keep the animation engines in a map of "pageNumber":engine to avoid making a new one each time we load a page
> ```

Sorry, but now you changed the code, the comment needs an update :-)
___
*[`src/shared/animation.ts` line 389 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9QPop4xNkmYqYdmvm:-OX9QPop4xNkmYqYdmvn:bowby14) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L389)):*
> ```XML
>     }
> 
>     // Adjust the wrap div for a change of orientation. The name is slightly obsolete
> ```

Not a big deal, but 'wrap div' has changed name. Maybe check if it needs to change in any other comments, too?
___
*[`src/shared/animation.ts` line 602 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9RqpW3am4TECgeTSh:-OX9RqpX9wGwHfrS4O1m:bhmoqp) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L602)):*
> ```XML
>     }
> 
>     public play() {
> ```

Should we call this 'resume'? Since it doesn't reset totalElapsedTime, it won't necessarily play anything unless there's been a previous call to startAnimation, and there could be an expectation that 'play' will start from the beginning if we haven't paused.
___
*[`src/shared/animation.ts` line 616 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9THEzEKKa92uC5FAf:-OX9THEzEKKa92uC5FAg:bih908e) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L616)):*
> ```XML
> 
>     public endAnimation() {
>         this.totalElapsedTime = this.duration + 1;
> ```

It vaguely worries me that this animation engine, which we're about to replace with a new one typically, will still receive another animation frame callback. Feels like continuing to use a disposed object. With the current implementation of advanceAnimation, that won't hurt anything, and I don't see a clear way to improve it, but you might. At least, I think it's worth commenting something like "There may be one more call to advanceAnimation from requestAnimationFrame, but it will see the total time has elapsed and so not do anything".
One other thought: as things stand, the animation will be left in whatever state it is when this is called. I don't suppose there's any reason we'd want to move the transform to the final state if it's not there already?

NJStriebel (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/386)* status: 3 of 5 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[`src/shared/animation.ts` line 66 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9PY7w-80Geu0Kf2fH:-OX9agsn-lub3AdwxHKv:b-896fix) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L66)):*

Sorry, but now you changed the code, the comment needs an update :-)
</blockquote></details>

Done.
___
*[`src/shared/animation.ts` line 389 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9QPop4xNkmYqYdmvm:-OX9awrE9JX4nVoepR8o:b-896fix) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L389)):*

Not a big deal, but 'wrap div' has changed name. Maybe check if it needs to change in any other comments, too?
</blockquote></details>

Done.
___
*[`src/shared/animation.ts` line 602 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9RqpW3am4TECgeTSh:-OX9ayBy8bNpuA7vca48:b-y5u8sv) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L602)):*

Should we call this 'resume'? Since it doesn't reset totalElapsedTime, it won't necessarily play anything unless there's been a previous call to startAnimation, and there could be an expectation that 'play' will start from the beginning if we haven't paused.
</blockquote></details>

Done. That's a great idea.
___
*[`src/shared/animation.ts` line 616 at r3](https://reviewable.io/reviews/BloomBooks/bloom-player/386#-OX9THEzEKKa92uC5FAf:-OX9cDgL6A8y4GYgZT1Q:bdm35rt) ([raw file](https://github.com/BloomBooks/bloom-player/blob/90636ead1aaaa964e090f8191d368ef43c9ca385/src/shared/animation.ts#L616)):*

It vaguely worries me that this animation engine, which we're about to replace with a new one typically, will still receive another animation frame callback. Feels like continuing to use a disposed object. With the current implementation of advanceAnimation, that won't hurt anything, and I don't see a clear way to improve it, but you might. At least, I think it's worth commenting something like "There may be one more call to advanceAnimation from requestAnimationFrame, but it will see the total time has elapsed and so not do anything".
One other thought: as things stand, the animation will be left in whatever state it is when this is called. I don't suppose there's any reason we'd want to move the transform to the final state if it's not there already?
</blockquote></details>

Done. I added that comment and another one in advanceAnimation explaining the necessity of the duration check. I don't see any way the current implementation could cause a problem, and I think those two comments will prevent anyone from carelessly making changes that would cause a problem.

Currently, we only call endAnimation when we're transferring control to a new animation engine, so it doesn't particularly matter what frame endAnimation sets the canvas to. I think stopping on the current frame is reasonable enough, but if you disagree that's a pretty trivial change to implement.

JohnThomson (approved):
> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/386)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @NJStriebel)

PR #387, opened 2025-08-19
fix: use original page size if book has any overlays (BL-15140)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/387)

### DIFF
--- a/src/bloom-player-core.tsx
+++ b/src/bloom-player-core.tsx
@@ -254,6 +254,7 @@ export class BloomPlayerCore extends React.Component<IProps, IPlayerState> {
 
     private bookInfo: BookInfo = new BookInfo();
     private bookInteraction: BookInteraction = new BookInteraction();
+    private mustUseOriginalPageSize: boolean = false;
 
     private static currentPagePlayer: BloomPlayerCore;
     private indicesOfPagesWhereWeShouldPreserveDOMState: any = {};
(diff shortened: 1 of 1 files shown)


JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/387)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`src/bloom-player-core.tsx` line 1374 at r1](https://reviewable.io/reviews/BloomBooks/bloom-player/387#-OY3pdil6wEmvTumJwUs:-OY8UlNB1IId9zx2EFks:b14uzx4) ([raw file](https://github.com/BloomBooks/bloom-player/blob/c4c0928cd5160ef97005b410aa8d9ae9c5d51b3c/src/bloom-player-core.tsx#L1374)):*

Hm. Ok, I guess I got a little confused thinking it was only the case of an svg that we can't scale things down well to reposition things. But I guess we don't try to reposition things at all after publication.
</blockquote></details>

Yes, another plausible approach is to say that when we don't have an SVG, we assume that any overlays can be resized. Bloom Desktop has an algorithm for doing this which works pretty well for anything but text, and would often be OK for the few text boxes that don't result in SVGs. I don't know which is more desirable. The best option, of course, is for the author to position things the way he wants on a Device page size if he plans an electronic publication. Scaling the page makes for greater fidelity to the original; resizing typically makes use of more of the screen. Maybe a JohnH decision? I went for a quick, easy, and I thought safe fix.

PR #428, opened 2026-06-15
fix: improve audio loading

JohnThomson:
Bloom will try to preload the next audio segment while playing the previous one.
Bloom will not highlight a segment for audio until the audio is ready to play (especially non-initial segments)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/428)

### DIFF
--- a/src/activities/ActivityContext.ts
+++ b/src/activities/ActivityContext.ts
@@ -3,7 +3,12 @@ import {
     getPageData,
     storePageData,
 } from "../page-api";
-import { setDefaultSoundUrls, urlPrefix } from "../shared";
+import {
+    cleanupActivitySounds,
+    playCorrectOrWrongSound,
+    preloadSoundsForActivity,
+    setDefaultSoundUrls,
+} from "../shared";
 import rightAnswer from "./right_answer.mp3";
 import wrongAnswer from "./wrong_answer.mp3";
 
@@ -38,6 +43,7 @@ export class ActivityContext {
             this.fixViteSoundPath(rightAnswer),
             this.fixViteSoundPath(wrongAnswer),
         );
+        preloadSoundsForActivity(this.pageElement);
     }
 
     // Report a score that can be used for analytics. The caller can call this repeatedly without worrying
@@ -74,11 +80,11 @@ export class ActivityContext {
     }
 
     public playCorrect(evt: Event) {
-        this.playCorrectOrWrong(evt, rightAnswer, "data-correct-sound");
+        playCorrectOrWrongSound(this.pageElement, true);
     }
 
     public playWrong(evt: Event) {
-        this.playCorrectOrWrong(evt, wrongAnswer, "data-wrong-sound");
+        playCorrectOrWrongSound(this.pageElement, false);
     }
 
     fixViteSoundPath(path: string) {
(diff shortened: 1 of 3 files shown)


JohnThomson (commented):
> @JohnThomson made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/428)* status: 0 of 3 files reviewed, 1 unresolved discussion.

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/bloom-player/428#-OvBzyMb2ToURXbwfFSP:-OvBzyMcBgz2kq4zLJnh:bl9m551) (no related file):*
Devin is happy as of 57384b8866c45ae9c5e186c2c2a589aeea56f065

hatton:
That's _lot_ of code for what we're getting.  Do you have a sense of whether the AI overbuilt it or that really is what's required?

JohnThomson (commented):
Each bit of code seemed to make sense. There's code to preload sounds, which requires a different audio element if the main one is already playing the previous sound. Complicated by our cache-busting, which then has to be defeated because we do NOT want to bust the cache between preloading a sound into the extra player and playing it in the main one. Then Devin picked up that we were accumulating listeners...I'm not sure whether that was new. We added pre-loading for game correct/wrong sounds. We fiddled with event handlers to suppress the audio highlight until the sound starts playing. I didn't study it in great depth, searching for simplifications.
Whether it's required is harder to say. With my fast network connection, and perhaps low sensitivity to some kinds of imperfections, I didn't notice a problem...assuming I even found the right book. So I was just making some changes that seemed as if they might possibly help.

> @JohnThomson made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/428#-:-OvC19RoBH-wATYQLOZD:b-knbfkz)* status: 0 of 3 files reviewed, 1 unresolved discussion.

hatton:
Opus says

The commit bundles three separate concerns...

The core idea is right and not excessive: create the <audio> once at activity-prepare time, call .load(), keep the reference, .play() on click. That's the minimal correct fix for "first click has a delay."

**What's actually redundant is that it's implemented twice, in two different styles, because the two activity families take different code paths:**

ActivityContext (used by [SimpleCheckboxQuiz.ts:157](vscode-webview://1i07vutd7no50p845qg2ghaignd8s28panlk9sp2n3iuj86m3ep4/src/activities/domActivities/SimpleCheckboxQuiz.ts#L157) and [SimpleDomChoice.ts:88](vscode-webview://1i07vutd7no50p845qg2ghaignd8s28panlk9sp2n3iuj86m3ep4/src/activities/domActivities/SimpleDomChoice.ts#L88)) → preloads into instance fields correctSoundPlayer/wrongSoundPlayer, no DOM.
dragActivityRuntime (drag activities) → preloads into DOM <audio> elements tagged data-preload-sound, with dedup-by-DOM-query, a "consume by deleting the attribute" dance, bloom-ui class, and cleanup in undoPrepareActivity.
So now the correct/wrong URL-resolution logic lives in ~3 places (preloadActivitySounds, playCorrectOrWrong/fixViteSoundPath, and preloadSoundsForActivity), in two incompatible caching styles. That's the real smell — not the amount of work, but the duplication.

The dragActivityRuntime version is the heavier of the two because it has to slot into the existing playSound() that's shared with the Bloom toolbox and needs the DOM append for relative-URL resolution (per the existing comment). The querying-by-attribute and attribute-deletion to mark "consumed" is the most baroque part and could be a simple module-level Map<url, HTMLAudioElement>.

**Where it could genuinely shrink**
Collapse the two game-sound preloaders into one shared helper both families call.
Possibly lean on the browser HTTP cache instead of holding element references: fire a .load() of the sound URLs at prepare time, then new Audio(url) on click hits cache and plays instantly — if you **drop the ?nocache= buster** for these sounds. That's what the narration code is effectively doing manually (storing the exact URL so the cache hits).

JohnThomson (commented):
I had claude attempt to reduce the duplication.

> @JohnThomson made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/428#-:-OvL-Dzq4i5x3xVbXFt_:b91iqtk)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on JohnThomson).

PR #434, opened 2026-07-16
chore: upgrade TypeScript to 7.0 (BL-16520)

andrew-polk:
Upgrades bloom-player to TypeScript 7.0.2, the native (Go-based) compiler.

No code changes were needed — the BL-16559 work (target es2018 plus the accumulated type-error fixes) already satisfies TS 7. `tsc --noEmit` here now runs on the native compiler (~2× faster even on this small repo, with much bigger wins on large ones).

Two companion changes make the toolchain work, because TS 7's npm package no longer ships the JavaScript compiler API that tooling embeds:

- `vite-plugin-dts` 5.0.0 → 5.0.3, which knows how to fall back when the JS API is absent;
- add `@typescript/typescript6`, Microsoft's official JS-API fallback package, used only by declaration-file generation during `build:sharedlib`.

Verified: `tsc --noEmit` clean; full vitest suite passes; both builds succeed; and all generated `.d.ts` files are **byte-identical** to the TS 6 output, so the published type surface is unchanged. (vite/vitest/tsx/storybook transform TS via esbuild and never touch the compiler API.)

The one transition-period caveat: if source ever uses TS-7-only syntax, declaration generation (still driven by the TS 6 engine) would fail loudly at build time until the plugin ecosystem adopts the native API — at which point `@typescript/typescript6` can be dropped.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16520 (multi-repo ticket; this PR covers bloom-player)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/434)


andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-07-16 up to commit bee69ff323f0e8ff4f9755f2c31d2bfae74ad6de. No bugs and no investigate-level flags (0/0). One informational observation, not mirrored as a review thread: the TS 6 fallback for declaration generation relies on module resolution finding @typescript/typescript6 — addressed by design, since we declare it as a direct, exactly-pinned devDependency. Devin review page: https://devinreview.com/BloomBooks/bloom-player/pull/434

andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-07-16 up to commit b45a8b4b5fadee65c154f998a6b2952d086f7c9d (the rebase onto alpha + the .test.tsx tsconfig-exclude fix). No bugs and no investigate-level flags (0/0). Two informational observations, not actionable: the TS 6 declaration-generation fallback resolving via a direct pinned devDependency (by design), and a note that .test.tsx files are now excluded from the tsc program (the deliberate fix — test files are checked by vitest, matching the existing **/*.test.ts exclusion). Devin review page: https://devinreview.com/BloomBooks/bloom-player/pull/434

PR #435, opened 2026-07-16
chore: fix `pnpm storybook` under TypeScript 7.0

andrew-polk:
## Problem

`pnpm storybook` crashed on startup:

```
Failed to load preset: @storybook/react/dist/preset.js
TypeError: Cannot read properties of undefined (reading 'React')
  at react-docgen-typescript/lib/parser.js:27
```

## Cause

The recent TypeScript 7.0 upgrade (BL-16520). In TS 7 the `typescript` package's main entry maps to `lib/version.cjs`, which exports only `{ version, versionMajorMinor }` — the classic enum objects (`JsxEmit`, `ModuleKind`, `ScriptTarget`) are no longer runtime values. `@storybook/react`'s preset does an **unconditional top-level import** of `react-docgen-typescript`, whose `parser.js` reads `ts.JsxEmit.React` at module load — so it throws the moment the preset is imported, before any config is consulted.

## Fix

Two complementary changes:

- **`.storybook/main.ts`** — `typescript: { reactDocgen: false }`. We don't use the generated prop tables (the stories are iframe-based BloomPlayer stories), so docgen is turned off entirely. This is our actual intent.
- **`patches/react-docgen-typescript@2.4.0.patch`** — because the preset imports the module regardless of the config flag, guard the three module-level enum reads with their historical numeric fallbacks (`JsxEmit.React`→2, `ModuleKind.CommonJS`→1, `ScriptTarget.Latest`→99) so the now-unused module survives import. Registered via pnpm `patchedDependencies` in `pnpm-workspace.yaml`.

Config alone is not sufficient (verified) — the import runs at preset-load time. The patch is dead code we only need to survive import; it can be dropped once Storybook makes that import lazy or `react-docgen-typescript` ships TS-7 support.

## Verification

- `pnpm storybook` boots cleanly from a fresh `pnpm install`: preview serves (HTTP 200), all 34 stories load, no preset/docgen errors.
- `pnpm exec tsc --noEmit` passes.
- `pnpm exec vitest run` — 120/120 tests pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/435)

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -17,6 +17,12 @@ const config: StorybookConfig = {
         name: "@storybook/react-vite",
         options: {},
     },
+    typescript: {
+        // We don't use the auto-generated prop tables, and react-docgen-typescript
+        // crashes under TypeScript 7.0 (it reads enum values the TS 7 package no
+        // longer exposes at runtime). Turn docgen off entirely.
+        reactDocgen: false,
+    },
     viteFinal: (config, options) => {
         const idToBookName = {
             "2e492eb1-bcc5-4b2b-b756-6cda33e1eee4": "multibook-index",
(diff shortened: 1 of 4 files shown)


andrew-polk:
[Claude Opus 4.8] Consulted Devin on 2026-07-16 12:30 USMST up to commit `34307f9`. Result: **0 Bugs, 0 Investigate flags**. Two Informational notes only, both confirming the change (patch registration consistent with the lockfile; enum fallback values `2/1/99` historically correct). Nothing to action.

PR #437, opened 2026-07-17
fix: replace BloomPlayerCore static back-channels with a registry, clean up listeners (BL-16554)

andrew-polk:
React modernization **Phase 2**, stacked on `phase1-shrink-core`.

Replaces `BloomPlayerCore`'s mutable statics (`currentPagePlayer`, `currentPage`, `currentPageIndex`, `currentPageHasVideo`, `getCurrentPage`, `storeVideoAnalytics`) with instance members. Non-React modules (`video.ts` analytics, `page-api.ts`) now reach the live player through an explicit `ICurrentPlayer` registry (`currentPlayer.ts`) that the player registers with on mount and — unlike the old static, which was never cleared — unregisters from on unmount. Also removes an unused `bloom-player-core` import from `music.ts` (breaking an import cycle).

Fixes real listener leaks: the window focus/blur, document keydown, and the capture-phase pointerdown/dragstart handlers were added with anonymous functions and never (or ineffectively) removed; `componentWillUnmount` now removes the exact handler instances.

**Base:** targets `phase1-shrink-core` (the Phase 1 branch), not `alpha` — this is the second link in the stacked refactor series.

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16554

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/437)

PR #439, opened 2026-07-22
fix: play all draggable answer videos on Show Correct, and show them on iOS (BL-16146)

JohnThomson:
Fixes the two remaining BL-16146 problems with draggable ("answer") videos in drag games:

**1. Show Correct only played the first answer video (timing-dependent).**
`showCorrectOrWrongItems` plays the correct/wrong feedback media in a callback that fires when the correct/wrong *sound* finishes. If the user clicked Show Correct before the sound ended, that late callback still called `playAllVideo`, bumping the playback generation counter and silently cancelling the solution-video sequence Show Correct had just started. The fix skips the feedback playback when the page has already left the state it was queued for.

Verified with an automated Playwright loop against the repro book: before the fix, clicking Show Correct within ~1s of Check failed every time; after, 48/48 runs across click delays of 150ms–2.2s played all three answer videos.

**2. Answer videos were invisible on iOS.**
The first-frame primer (play-then-pause to get a frame painted over the transparent poster) waited for `loadeddata` and played unmuted. On iOS Safari, video data isn't fetched until something plays (so `loadeddata` never fires) and gestureless unmuted `play()` is rejected — so the primer never ran and the poster never came off. The primer now plays muted immediately (the `play()` call itself is what makes iOS load the data), pauses on the first composited frame, restores the muted state, and removes the transparent poster; a 3s give-up fallback removes the poster regardless so a video can never stay invisible. Real-playback initiators (both `playAllVideo` implementations and the draggable-tap handler) deterministically cancel a pending priming attempt so it can't mute or pause playback that takes the video over.

Also adapts the five analytics tests that the phase2-remove-statics refactor (#437) broke on alpha (they still called the removed `BloomPlayerCore` statics).

**Note:** the iOS behavior still needs confirmation on a real device — desktop browsers don't emulate iOS media policies. One thing to watch there: whether videos 2 and 3 of the Show Correct sequence play (their `play()` calls happen in `ended` handlers, outside the tap's gesture context).

Ref: https://issues.bloomlibrary.org/youtrack/issue/BL-16146

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/439)

### DIFF
--- a/src/shared/narration.ts
+++ b/src/shared/narration.ts
@@ -1401,6 +1401,18 @@ export function isTransientVideoPlayFailure(reason: any): boolean {
     return message.includes("interrupted by a call to pause()");
 }
 
+// Cancel functions for in-progress first-frame priming attempts (see
+// showVideoFirstFrameWhenReady below), keyed by the video being primed.
+const firstFramePrimingCancels = new WeakMap<HTMLVideoElement, () => void>();
+
+// Anything about to start real playback of a video should call this first:
+// if a first-frame priming attempt is in progress on the video, it is
+// cancelled (restoring the video's muted state and removing the transparent
+// poster) so the primer can't mute or pause the real playback.
+export function cancelVideoFirstFramePriming(video: HTMLVideoElement) {
+    firstFramePrimingCancels.get(video)?.();
+}
+
 // Attempt to show a video's first frame by briefly starting muted playback and
 // then pausing. We mute while priming because gestureless unmuted play() is
 // rejected by autoplay policies (iOS especially, where a past tap elsewhere on
@@ -1430,16 +1442,22 @@ export function showVideoFirstFrameWhenReady(
             return;
         }
         done = true;
+        firstFramePrimingCancels.delete(video);
         window.clearTimeout(giveUpTimeout);
         video.removeEventListener("playing", playingListener);
         video.muted = wasMuted;
     };
-    // If playAllVideo has meanwhile taken this video for real playback, we
-    // must neither pause it nor leave it muted.
-    const sequencerOwnsVideo = () => activePlayAllVideoElement === video;
+    // Let real-playback initiators cancel this priming attempt so it can't
+    // mute or pause the playback they are about to start.
+    firstFramePrimingCancels.set(video, () => {
+        finish();
+        // Real playback is starting; make sure our transparent poster can't
+        // hide it.
+        video.removeAttribute("poster");
+    });
     const playingListener = () => {
         hideVideoAutoplayBlockedHint(video);
-        if (!shouldPauseAfterPlaying() || sequencerOwnsVideo()) {
+        if (!shouldPauseAfterPlaying()) {
             // Real playback has taken over; restore audio and get out of the way.
             finish();
             video.removeAttribute("poster");
@@ -1449,7 +1467,7 @@ export function showVideoFirstFrameWhenReady(
             if (done) {
                 return; // we already gave up or something else took over
             }
-            if (shouldPauseAfterPlaying() && !sequencerOwnsVideo()) {
+            if (shouldPauseAfterPlaying()) {
                 video.pause();
             }
             finish();
@@ -1481,8 +1499,7 @@ export function showVideoFirstFrameWhenReady(
     // audio, and remove the transparent poster so the browser can show its own
     // first frame when it has one, rather than leaving the video invisible.
     giveUpTimeout = window.setTimeout(() => {
-        const pauseWanted = shouldPauseAfterPlaying() && !sequencerOwnsVideo();
-        if (pauseWanted && !video.paused) {
+        if (shouldPauseAfterPlaying() && !video.paused) {
             video.pause();
         }
         finish();
@@ -1570,6 +1587,9 @@ function playAllVideoInternal(
         hideVideoError(video);
         hideVideoAutoplayBlockedHint(video);
         setCurrentPlaybackMode(PlaybackMode.VideoPlaying);
+        // This is real playback; a pending first-frame priming attempt must
+        // not mute or pause it.
+        cancelVideoFirstFramePriming(video);
         // Always play each queued video from the beginning.
         // Without this, a previously played element may remain at end-of-stream
         // and fail to raise the expected ended event for sequencing.
--- a/src/shared/narration.ts
+++ b/src/shared/narration.ts
@@ -1432,6 +1432,12 @@ export function showVideoFirstFrameWhenReady(
     if (!video.paused) {
         return;
     }
+    // If an earlier priming attempt on this video is still armed (e.g. the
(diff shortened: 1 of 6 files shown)


src/shared/narration.ts
```diff
@@ -1401,94 +1401,134 @@ export function isTransientVideoPlayFailure(reason: any): boolean {
...
+    };
+    // Let real-playback initiators cancel this priming attempt so it can't
+    // mute or pause the playback they are about to start.
+    const cancel = () => {
+        finish();
+        // Real playback is starting; make sure our transparent poster can't
+        // hide it.
+        video.removeAttribute("poster");
+    };
+    firstFramePrimingCancels.set(video, cancel);
```
JohnThomson, line 1468: [Devin] **Investigate**: Concurrent re-priming of the same video can orphan the cancel registration

Devin (on commit 9f8a8a9): showVideoFirstFrameWhenReady stores a single cancel closure per video in the firstFramePrimingCancels WeakMap, and finish() unconditionally deleted the map entry. If the same video element were primed twice while the first attempt was still pending (e.g. re-entering the same swiper-slide page via undoPrepareActivity/prepareActivity within the 3s window), the second call would overwrite the map entry, and when the FIRST attempt's finish() later ran it would delete the SECOND attempt's cancel closure. A subsequent real-playback initiator's cancelVideoFirstFramePriming would then find nothing, so the still-live second primer could mute/pause the real playback.

JohnThomson: [Claude Fable 5] Fixed in 35c9a09: a new priming attempt now cancels any still-armed earlier attempt on the same video (so there is only ever one live attempt per video, and the muted state nests correctly), and cleanup deregisters only its own cancel function. Regression test added ("re-priming a video cancels the earlier attempt instead of stacking"). Resolving.

src/shared/narration.ts
```diff
@@ -1401,94 +1401,134 @@ export function isTransientVideoPlayFailure(reason: any): boolean {
...
+// then pausing. We mute while priming because gestureless unmuted play() is
+// rejected by autoplay policies (iOS especially, where a past tap elsewhere on
+// the page grants nothing). Calling play() is also what forces browsers that
+// don't preload video data (again, iOS Safari) to fetch and decode the first
+// frame, so we must not wait for loadeddata first: on iOS it typically never
+// fires until something plays. (BL-16146)
+// The callback allows callers to decide at the moment playback starts whether
+// pausing is still desired (for example, page-level logic may stop pausing
+// after initial load).
 export function showVideoFirstFrameWhenReady(
```
JohnThomson, line 1426: [Devin] **Investigate**: iOS-specific media policy behavior remains unverified on real devices

The primer rework hinges on iOS Safari semantics that desktop browsers and jsdom cannot emulate: that a muted gestureless play() is what forces data loading and is not rejected, and that requestVideoFrameCallback/playing fire as assumed. The PR description itself flags that this needs confirmation on a real device, and specifically that videos 2/3 of the Show Correct sequence (play() invoked from ended handlers, outside the tap gesture) may still be blocked. This is not a code defect I can confirm, but the branch's correctness on iOS is unverified and warrants device testing before relying on it in production.

JohnThomson: [Claude Fable 5] Agreed — this is a verification gap, not a code defect. It is deliberately called out in the PR description, and the QA test-ideas comment on BL-16146 includes a specific iOS device checklist (first frames visible, tap-to-play inline, and whether all three Show Correct videos play). Device testing is on the pre-merge checklist; resolving the thread since the tracking lives there.

JohnThomson:
[Claude Fable 5] Consulted Devin on 2026-07-22 up to commit 35c9a098427255420d3a8d8b9bb0d775e9a54f60 (two rounds). Round 1 (9f8a8a9): 0 bugs, 1 Investigate flag — fixed in 35c9a09, thread resolved (https://github.com/BloomBooks/bloom-player/pull/439#discussion_r3633138468). Round 2 (35c9a09): 0 bugs; the round-1 flag is confirmed Resolved by Devin; 1 new Investigate flag noting iOS behavior needs real-device verification — acknowledged and tracked in the QA test plan on BL-16146 (https://github.com/BloomBooks/bloom-player/pull/439#discussion_r3633178157). Informational items skipped. No PR CI is configured on this repo; local gates (tsc, vitest 153/153, Playwright e2e 19/19) are green at 35c9a09.

JohnThomson:
[Claude Fable 5] iOS device verification done (2026-07-22): served this branch's built player + the BL-16146 repro book over a hotspot to a real iOS device. All three concerns check out — the draggable answer videos show their first frames, tapping plays inline with sound, and Check → Show Correct plays **all three** answer videos in sequence (including videos 2 and 3, whose play() calls run outside the tap gesture — the concern raised in the PR description and Devin's round-2 flag). The verification gap noted in https://github.com/BloomBooks/bloom-player/pull/439#discussion_r3633178157 is closed.

StephenMcConnel (approved):
> @StephenMcConnel reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/439)* status: all files reviewed, 2 unresolved discussions.

StephenMcConnel (approved):
> @StephenMcConnel resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/439)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on JohnThomson).

hatton:
@copilot resolve the merge conflicts in this pull request

hatton (changes requested):
> @hatton reviewed 2 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/439)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on JohnThomson and StephenMcConnel).

___
*[`src/shared/narration.ts` line 1416 at r2](https://reviewable.io/reviews/BloomBooks/bloom-player/439#-Oy_Xiew5GAQhjoL9i1h:-Oy_Xiew5GAQhjoL9i1i:b-lox6do) ([raw file](https://github.com/BloomBooks/bloom-player/blob/6e586488dae7e4a442ba34669aedde5cef05cf3f/src/shared/narration.ts#L1416)):*
> ```XML
> }
> 
> // Attempt to show a video's first frame by briefly starting muted playback and
> ```

So I see how this works for sign language but what if the video was instructional, like with sound?

hatton (commented):
> @hatton resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/bloom-player/439)* status: 5 of 6 files reviewed, all discussions resolved (waiting on StephenMcConnel).

